Issue #25, opened 2016-03-13
Cannot understand how voting procedure works


danielepercivaldi:
How the poll appears:
![cattura](https://cloud.githubusercontent.com/assets/8720648/13731849/fe32de38-e973-11e5-9050-deee1e4b9e7e.JPG)
If I press on the "yellow area" it becomes:
![cattura](https://cloud.githubusercontent.com/assets/8720648/13731859/26759a66-e974-11e5-94d3-13ad9645fd2e.JPG)
and again
![cattura](https://cloud.githubusercontent.com/assets/8720648/13731868/5839fc4a-e974-11e5-8108-44c306b6d2cd.JPG)
If I continue 
![cattura](https://cloud.githubusercontent.com/assets/8720648/13731874/8d7691de-e974-11e5-9553-d13cba1cf112.JPG)

Please, explain to me. Is it rigth?

v1r0x:
The colored area on the left of the vote cells is a "toggle-all" cell. It should turn all cells to the color of the "toggle-all" color. This cell should work as all other cells.

But you are right. There should be some explanation what it does. Maybe @jancborchardt has an idea? (Sorry for mention you ;))

danielepercivaldi:
Ok, I see: it works only for event polls.
In text polls toggle-all does not work and you cannot vote single item (the pointer change but when you clic nothing happens).

v1r0x:
This should be fixed in #27 as well.

jancborchardt:
@v1r0x just as driveby feedback: There’s a bit too much bold in the interface. :) And also the dark blue should not be used anywhere except primary action buttons.

PR #27, opened 2016-03-14
Fix text votes

### LINKED ISSUE #24: Internal server error happens when voting
cattura

v1r0x:
fixes #24, #25


danielepercivaldi:
No more internal server error when voting.
You can vote single text item.

PR #28, opened 2016-03-14
fix textarea creation

### LINKED ISSUE #26: When you edit a poll the description field is empty
When you edit a poll that has got a description, description field becomes empty (both texts and events polls).

v1r0x:
Loads the description, if set. Fixes #26


danielepercivaldi:
It's ok.

PR #29, opened 2016-03-14
correctly show texts on edit

### LINKED ISSUE #23: Text based polls editing does not show texts
If you define a poll like this:
cattura

When you edit it you cannot see its own texts:
cattura

v1r0x:
fixes #23


danielepercivaldi:
Now text items are shown in editing polls.

PR #39, opened 2016-03-15
redirect to login before error page

### LINKED ISSUE #34: Poll link while logged out results in confusing error message
If not logged in the poll link should show loggin mask instead telling the poll does not exist or you lack rights

Error message:

> Fehler
> Ihnen fehlt die Berechtigung zur Anzeige dieser Abstimmung oder diese Abstimmung existiert nicht.

v1r0x:
fixes #34, what do you think @dekuenstle ?


dekuenstle:
This should do it. Could you test it on a machine?

v1r0x:
Tested it and it works. If it is not a public poll and the user is not logged in, they have to log in and then the error appears (or the poll ;))

PR #40, opened 2016-03-15
avoid negative total values

### LINKED ISSUE #36: Poll initialization
When you enter a poll the first time and never voted:
cattura
I think it should instead be:
cattura
because if you vote a single item (-1 ?!?):
cattura

v1r0x:
fixes #36, @danielepercivaldi please review.


danielepercivaldi:
It's ok. Thanks.

Issue #67, opened 2016-05-24
Only show users from Group


stonerl:
When creating a new poll the app currently shows all users and not only the users which are part of the creators group. E.g:

Group1: User 1, User 2
Group2: User 2, User 3

When User 1 creates a poll, only Group 1 is listed but all Users. I don't think that it should be like this. With the current behavior all Users of the server are exposed to the creator. The app should respect the option `Restrict users to only share with users in their groups`.

v1r0x:
You are right. Will tweak this. Either restrict users to users of groups the current user is member of or try to get users based on the share settings (in the admin panel).

bodo1987:
As already mentioned in my pull request, I proposed a solution which only shows users of groups the current user is member of. For me, this seems to be a logic solution. However, I am not sure if this is the intended behavior in all cases. For instance, if you are admin and normal users don't have a group, you probably would like to have them in your list as well.
During writing of this comment, I came to the conclusion that this might not be the optimal solution. Nevertheless, it can serve as a starting point ;)

bodo1987:
OK .... #68 shows always users without any group now 

Looks better for me  

RFC

v1r0x:
Is this fixed for you @stonerl ?

stonerl:
yes

Issue #50, opened 2016-04-14
Just some groups and persons available to choose


scroom:
Hi,

thanks for your great app! Just installed Polls in our schools ownCloud instance. But then I've realized that only a few groups are shown if I click on choose (it seems that these are the groups I'm member of) and as well only some random users. We're using a LDAP backend for user authentication.
For our usecase it would be great if one could choose a group (a class or course) and create a poll for them. Normally everyone can see every group (and share for example files with them).

v1r0x:
It is intended that you only see groups you are a member of. Thought there wouldn't be a case where you want to create polls for other groups.
Do the "random users" have something in common? We use LDAP as well and don't have any problems with it. Maybe some group or admin connection between those users.
I hope I have some time this weekend to switch from *only user groups* to *all groups*

scroom:
In fact for us it would be the best to have a search field (similar to the search field in ownCloud itself to share files etc.) out of which one could add groups or persons to the poll.
It would prevent the problem that only random users are loaded in the user field (around 50 but we are having more than 900 users)

v1r0x:
I'll think about it and try to find a proper solution.

v1r0x:
@scroom It's been a while, but thanks to @bodo1987 the user/group pick list has been cleaned up. Does your problem still exist or has it been fixed with the new version? (Check out current master to see the result)

scroom:
Hey!

Sorry for responding so late. My problem is solved partially. No I can search for every registered user. But still I can just choose the groups I'm in. Would be great for our use case (a school) if we could register poll for classes (teachers are not part of the class und courses groups).

v1r0x:
So teachers are in different groups and no admins?
I think your use case is a very special one, thus I think it would be _wrong_ to enable this feature for all users. Nevertheless it would be good to find a solution for your particular problem.
I think we shouldn't expose all groups to all users or only expose them depending on the following admin settings:
- Allow sharing with groups
- Restrict users to only share with users in their groups
- Exclude groups from sharing

I don't know if it is possible to check for these settings, but if this would be a solution for you, I'd check this and try to fix it that way.
What's your opinion?

scroom:
Yes teachers and students are in different groups (in fact the situation in real life is emulated - teachers are not part of classes or courses but they are teaching them).

So your solution seems to fit this usecase. Would be really cool if we could use the polls app for a lot of use cases in the whole school (nearly 900 users).

v1r0x:
[...]

PR #68, opened 2016-05-27
changes in user and group pick list

bodo1987:
This should solve the issues #67, it least it works quit fine for me. Just try it out and merge it if you are happy with it. 

Cheers,
bodo

### DIFF
--- a/templates/create.tmpl.php
+++ b/templates/create.tmpl.php
@@ -79,7 +79,6 @@
             
             <div id="access_rights" class="row">
                 <div class="col-50">
-                    <!--<h3><?php p($l->t('Groups')); ?></h3> -->
                     <input type="text" class="live-search-box-group" placeholder="<?php p($l->t('Group search')); ?>" />
 		    <ul class="live-search-list-group">
                         <?php 
@@ -93,7 +92,6 @@
                     </ul>
                 </div>
                 <div class="col-50">
-	   	    <!--<h3><?php p($l->t('Users')); ?></h3> -->
 		    <input type="text" class="live-search-box-user" placeholder="<?php p($l->t('User search')); ?>" />
                     <ul class="live-search-list-user">
                     	<?php 
@@ -108,7 +106,10 @@
 			sort($users, SORT_NATURAL | SORT_FLAG_CASE );
                         foreach ($users as $user) : ?>
                             <li class="cl_user_item cl_access_item" id="user_<?php p($user); ?>" >
-                                <?php p($user); ?>
+				<?php p($userMgr->get($user)->getDisplayName()); ?>
+				<span id="sec_name">
+				<?php p($user); ?>
+				</span>
                             </li>
                         <?php endforeach; ?>
                     </ul>
--- a/templates/create.tmpl.php
+++ b/templates/create.tmpl.php
@@ -92,7 +92,13 @@
                 <div class="col-50">
                     <h3><?php p($l->t('Users')); ?></h3>
                     <ul>
-                        <?php $users = OC_Group::usersInGroups($groups); ?>
+                        <?php $all_groups = OC_Group::getGroups(); ?>
+			<?php $users = OC_User::getUsers(); ?>
+                        <?php $users_in_groups = OC_Group::usersInGroups($all_groups); ?>
+                        <?php $users_no_group = array_diff($users, $users_in_groups); ?>
+			<?php $users = OC_Group::usersInGroups($groups); ?>
+			<?php $users = array_merge($users , $users_no_group); ?>
+			<?php sort($users, SORT_NATURAL | SORT_FLAG_CASE ); ?>
                         <?php foreach ($users as $user) : ?>
                             <li class="cl_user_item cl_access_item" id="user_<?php p($user); ?>" >
                                 <?php p($user); ?>
--- a/templates/create.tmpl.php
+++ b/templates/create.tmpl.php
@@ -79,10 +79,13 @@
             
             <div id="access_rights" class="row">
                 <div class="col-50">
-                    <h3><?php p($l->t('Groups')); ?></h3>
-                    <ul>
-                        <?php $groups = OC_Group::getUserGroups($userId); ?>
-                        <?php foreach($groups as $gid) : ?>
+                    <!--<h3><?php p($l->t('Groups')); ?></h3> -->
+                    <input type="text" class="live-search-box-group" placeholder="<?php p($l->t('Group search')); ?>" />
+		    <ul class="live-search-list-group">
+                        <?php 
+			$groups = OC_Group::getUserGroups($userId);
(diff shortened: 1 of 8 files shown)


v1r0x:
thanks for your PR!
I'll have a look at it and merge it or let you know what's missing. For now: Maybe you could add a check if the current user is an admin and thus add all users.

bodo1987:
good idea!
done!

Cheers,
Bodo

bodo1987:
I just found #50 and implemented the search functions for users and groups. Even if I am not sure if all problems regarding the random population of the pick list are solved (I couldn't reprdouce this issue), it should still provide a better way of navigating through a large pick list now.

RFC

Cheers,
Bodo

v1r0x:
I just tested your code and looks good so far! One last thing: On my server I use rather cryptic user names. Would be great if the search would search for user names **and** display names. I think that's the way the sharing app does. Could you add this?

stonerl:
I'd like to add my 2ct. The list should also contain the display names instead of the user names when available.

bodo1987:
OK - "display names" are displayed as well. Maybe, this would look better when the names in brackets are displayed in a slightly different color (in grey) and the UID in bold letters.

But this has to be done by somebody else. My css skills are negligible ;).

v1r0x:
I'd recommend to only show display name and show user name on hover as `alt`.

bodo1987:
OK OK ... not really what u suggested but I think it is also a nice solution 

And I finally learned something about CSS 

RFC
Bodo

templates/create.tmpl.php
```diff
@@ -108,10 +108,10 @@
 			sort($users, SORT_NATURAL | SORT_FLAG_CASE );
                         foreach ($users as $user) : ?>
                             <li class="cl_user_item cl_access_item" id="user_<?php p($user); ?>" >
-				<?php p($user); ?>
-				(
 				<?php p(OC_User::getDisplayName($user)); ?>
```
v1r0x: `OC_User::getDisplayName($user)` is [deprecated](https://doc.owncloud.org/api/classes/OCP.User.html#method_getDisplayName). You should use `$userMgr->get($user)->getDisplayName()` instead.

bodo1987:
OK done

v1r0x:
Thanks for the changes!
I'm not completely happy with the current solution. We should either always display the real user name or only display it on hover or something like that.
Instead of
```php
<?php p($user); ?>
```
I'd propose something like this:
```php
<span id="sec_name"><?php p($user); ?></span>
```
and in the css something like this
```css
#sec_name {
    /*remove the display: hidden; value*/
    Element {
    font-style: italic;
    color: gray;
    font-size: 0.9em;
}
```

bodo1987:
I'm not so sure about this. Currently think about two cases:
The first one, somebody doesn't care about diisplay names. This would lead to same user and display names and whatever he search for he will find the correct user.
The other case is that somebody has numbered or random user names in the style user1/user2/ ... . In this case he probably want to search or pick the real names which are the display names. However, it may be the case that he only have an account ID – for some debug reasons or so. For this rare cases, he will search for the UserID and only under this circumstances it makes sense to show it. But in the normal (search for display name) case, he should not be bothered by cryptic names in the list.

v1r0x:
I think it is almost impossible to find a solution which fits all users. I came up with two new ideas.
1. Add a toggle to show display or user names. Imho this is too much effort for a not complete perfect solution
2. Use display name and use username as fallback if no display name is set. This is how it is currently done on the voting page. Imho the best solution/trade off for now. We should merge this as a first solution and maybe open a new issue to find a better solution. But with that we could have a working search functionality and users can report what the think about it.

What do you think?

bodo1987:
By default, the display name is the same as the user ID if you create a new user. Furthermore, I am not sure if it is possible to have no display name? For me, it doesn't work to delete the whole display name. If it is really not possible, then I wouldn’t see the need to check for empty fields. Besides, it would be relatively complex to check this and to handle all the exceptions.
My idea is to keep the current solution. At least if we can't find any use cases where it would be disadvantageous.

v1r0x:
I thought the `getDisplayName()` returns `null` if the display name is not set, but it returns the username, so you are right: no need to check it.
I'm not that happy with the current solution, because it adds/removes names "randomly". And I think that's one reason why OC has display names, so I would use this solution. And that's the way I did it somewhere else in the code.

bodo1987:
What do you mean by add/remove names "randomly"?  Can you give an example for this - maybe I have still a bug in the implementation. 

Besides this issue I am sorry, but I can't really follow your argumentation. 
I use the display names as default for the pick list - as you requested. However, the user is also free to search for a UserID. If he is doing so, then the userID is displayed in brackets. In my point of view a fair solution. 

And how do you manage to delete the whole display name? It doesn't work via GUI, at least not in my installation.

bodo1987:
OK - I checked again and you were right. I could find some cases where the UID was still displayed even if the search sting didn't match the UID anymore. 
Now the behavior should really be as described already above. 
By default only the display name is displayed.
Only if the search sting matches the UID, the latter is displayed as well.

Cheers,
Bodo

v1r0x:
Didn't talk about a bug, but nice that it was helpful finding one :wink: 
What I meant with "randomly" was the intended behaviour. Displaying the user name if the search term matches doesn't give the user any new information. The user can imagine that (if the search term doesn't match the display name) it matches the real user name. Displaying and removing the user name while typing may detract the user instead giving helpful information.

bodo1987:
Tonight, I got another idea. I could check if the UID and the display name are the same. If it is so, then the UID will never be displayed. So, we could remove the number of UIDs displayed and also, they would be limited to cases where they really add some information.
I would also like to get another opinion - what do you prefer stonerl?

stonerl:
@bodo1987 The standard behavior in ownCloud is to only show the display name. Since ownCloud does not allow empty display names there are only 2 cases to take care of; Either the username is equal to the display name or the username and display name is different.

When you share a file in ownCloud you can either enter the display name or the user name, but the UI will always show the display name only. So I would vote for this behavior since this is what users might expect...

bodo1987:
OK it would be most easy to implement it like this. I did it so far in my way in order to NOT confuse the user. I considered it misleading if you find a name which doesn't match at all to the search string.

However, I change it to the proposed behavior.

bodo1987:
done

I still like the previous version more. But I would say it is up to v1r0x to decide which version to merge.

Cheers,
Bodo

v1r0x:
Thanks for all the effort @bodo1987 really appreciate your help! and thanks for your thoughts @stonerl . I like the new version more.
One last thing: Could you squash some of the commits? E.g. the last 6 commits (starting from ea42600 to 03804ea). Thanks!

bodo1987:
done

v1r0x:
Great, thanks!

PR #69, opened 2016-06-03
some eye_candy

bodo1987:
I just played around with me new developed css skills ;D.

This is the outcome - have a look on it. In my opinion it is a bit more structured then before and for me it just looks better ;).

Cheers,
Bodo


v1r0x:
Thanks for the contribution!
IMO the current version looks cleaner and I like the colors more.

PR #77, opened 2016-08-09
augmented version to 0.7.1.1

joergmschulz:
This indicates we have a small change in the javascript area.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -3,7 +3,7 @@
    <id>polls</id>
    <name>Polls</name>
    <description>event schedule polls</description>
-   <version>0.7.1</version>
+   <version>0.7.1.1</version>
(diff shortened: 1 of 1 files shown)


v1r0x:
I'd like to stay with the x.y.z system. I hope I have some free time this week to finally fix some bugs and release a 0.7.2 version.

joergmschulz:
This is even better - I didn't dare to hope you can still spare time for polls (seeing things go slow here). Would you mind rising the max version for NC 10? From what I see in my NC10 beta, polls work just fine.

v1r0x:
Since I release this app still on the owncloud apps page I'd test it with oc instead of nc. I switched to nc myself, but I won't release it as nc app until the public release of the nc appstore.

PR #81, opened 2016-08-28
make current selected access type clickable

### LINKED ISSUE #47: You cannot enter users/group selection if "Select..." option is already set.
If "Select..." is already set in order to enter users/group selection window you must change option and then back to "Select ..." option.

v1r0x:
fix #47 please review @danielepercivaldi


danielepercivaldi:
Hi!
It works.

PR #90, opened 2016-10-21
visualize remove possibility

### LINKED ISSUE #87: UI: List of chosen users should have buttons to delete from list
As already mentioned here: https://github.com/v1r0x/polls/pull/80 i would be great to have something like a red cross to delete a user from the list of chosen users for a poll. Otherwise it might be unclear how to delete a user from a list (you have to click on the username up to now which is isn't intuitive).

v1r0x:
fix #87 @scroom please review

### DIFF
--- a/css/create.css
+++ b/css/create.css
@@ -33,8 +33,9 @@
 }
 
 .cl_access_item.selected:hover {
-    background-color: #337AB7;
+    background-color: #FF6F6F;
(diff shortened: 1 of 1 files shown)


scroom:
Sorry for responding so late. I have been/I am ill. This fix seems to solve the problem I've described. Really nice!

PR #104, opened 2016-11-14
Update UI

amittel:
Updated the UI with menu and breadcrumbs.

Open for discussion.


v1r0x:
Looks nice!
Feel free to mention me if you need support or if you're ready to merge. :)

amittel:
Thanks!
Right now it's just try & error.
I'm searching all over the place to get the right css classes and how stuff is done.

amittel:
@v1r0x I was thinking about the menu approach. Maybe it's better to do it like the files app.
Leaving the table and having the menu to only show linked, shared, favorites polls. Because if we have the menu and the table infos would be double and an empty index page isn't looking good.

v1r0x:
Sounds good. I'd keep the table and only show participated/commented, shared, etc polls in the menu.

v1r0x:
Hey @amittel,
any news on this? Really looking forward for the new UI!

amittel:
Hi!
Yes, working on it. Setting up a VM to work on it local.
Had no chance during holidays.

amittel:
Still not sure about the menu. Since all polls are shareable, it makes no sense.
Maybe we go for the access rights. But if we make the table sortable, it's also redundant...

The other thing is the header. We already got the name in the breadcrumb. Ditching the header where should the description go?

Todo:

- Make comments streamline
- Get vote table color from theming
- Hover icon on index table not quite right
- Better divider for voted squares

v1r0x:
Sorry for my late response.

> The other thing is the header. We already got the name in the breadcrumb. Ditching the header where should the description go?

Is this a real question? If so, could you post a screenshot with the layout?

> Create overlay, when deleting or sharing poll (http://t4t5.github.io/sweetalert/ or http://vodkabears.github.io/remodal/)

There should be a modal class in oc/nc, maybe this is sufficient?

If you need any help please contact me. I really like the new look! 😄

dartcafe:
Still nothing heard from @amittel. I ported some of his improvements in #151. @v1r0x Do you think we should close this PR?

v1r0x:
Will have a look at this one and your branch later. But I think the best would be to merge your changes and then close this one. If something from here is missing in your changes we can add them in a new PR.

PR #105, opened 2016-11-15
don't fetch hidden polls

### LINKED ISSUE #101: No difference between hidden and open poll?
In fact I tried to find out what the difference between hidden and open polls are, but have found none.

For both cases users can access the poll if they know the exact link and registered users can see both of them as well (there is as well no difference to the case registered users)

v1r0x:
fix #101 
@scroom does this fix the problem for you?


scroom:
Yes, seems to work as expected! Great!

PR #106, opened 2016-11-15
add extended anon type, replace username with random id

### LINKED ISSUE #100: Anonymous has no effect for polls admin
If I use the Anonymous-mode I've expected that I, as the one who created the poll, couldn't see who voted for what. But the opposite is the case: I can see who voted for what.

Maybe it would be a good idea to have an advanced Anonymous-mode. So no-one could be blamed for its choices.

v1r0x:
fix #100 
@scroom what do you think about it? Only problem I see is that users could vote several times since the name is replaced by a random id

### DIFF
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -62,6 +62,11 @@
                 <type>integer</type>
                 <default>0</default>
             </field>
+            <field>
+                <name>full_anonymous</name>
+                <type>integer</type>
+                <default>0</default>
+            </field>
 		</declaration>
 	</table>
     <table>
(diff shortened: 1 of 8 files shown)


scroom:
Isn't it possible to prevent voting several times (maybe it is sufficient to just hide the username (and maybe replace it by a random number or something else like user1, user2 ....?)

v1r0x:
Was my first idea, but this would give admins still the ability to have a look at the votes in the database. Maybe this is too much _security_?

v1r0x:
What's your opinion? Should double votes be allowed and admin users can not see who voted for what (even in the database) or should double votes be prevented but admins users can still see who voted for what in the database?

scroom:
If you mean with admin, that really just the cloud admin (and not the one who created the poll) can have a look in the database I would prefer to prevent double votes.

PR #109, opened 2016-12-06
add version strings for NC 10.2 and 11

joergmschulz:
add nextcloud version info - tests with 11 went ok; 10.2 ongoing and
look good.


v1r0x:
thanks for your contribution!

PR #115, opened 2017-02-26
add extended anon type

### LINKED ISSUE #100: Anonymous has no effect for polls admin

v1r0x:
Fix #100 Had to create a new branch/pull request. This is an updated version of #106 

Please review @scroom

### DIFF
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -62,6 +62,11 @@
                 <type>integer</type>
                 <default>0</default>
             </field>
+            <field>
+                <name>full_anonymous</name>
+                <type>integer</type>
+                <default>0</default>
+            </field>
 		</declaration>
 	</table>
     <table>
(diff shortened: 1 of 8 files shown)


scroom:
Should I test master? Or the branch new-full-anon? And it it a problem to test it with Nextcloud? (I no longer have OC installations but could create one if it is a problem)

v1r0x:
Please test new-full-anon. I don't think that this is a problem. I'm using OC and NC and polls works on both machines.

scroom:
Works great! (Just had to change the info.xml to make it run under NC) No more double votes and still anonymity!

v1r0x:
Perfect! Thanks for your help :smiley:

PR #116, opened 2017-02-26
Fix timezones

### LINKED ISSUE #97: Time zone issue?
With 0.7.2 there seems to be a time zone issue.

I select when creating a event poll times 9:00 10:00 etc. but in the actual poll they show up as 1:00, 2:00 etc.

Our server has a timezone set of UTC+8 (Manila) so it is likely that the app doesn't take this into account properly.

This might be related to some recent timezone issues in Owncloud/nextcloud also, but thanks for looking into it.

v1r0x:
Fix #97 

The timezones should now be in UTC in the database and the user's (browser's) timezone offset is added to the dates.

Please review @poVoq and @uyekt

### DIFF
--- a/controller/pagecontroller.php
+++ b/controller/pagecontroller.php
@@ -340,15 +340,10 @@ class PageController extends Controller {
             $ins = $this->eventMapper->insert($event);
             $poll_id = $ins->getId();
             sort($chosenDates);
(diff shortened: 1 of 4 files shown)


v1r0x:
Merge to finally finish the milestone. Please re-open the issue if it still doesn't work.

Issue #117, opened 2017-04-18
English translation has js errors


bogie:
en.js is showing a lot of console errors in chrome, I fixed it manually:

```
diff --git a/l10n/en.js b/l10n/en.js
index 99eaad7..fbc210b 100644
--- a/l10n/en.js
+++ b/l10n/en.js
@@ -63,14 +63,14 @@ OC.L10N.register(
     "You must enter at least a title for the new poll." : "You must enter at least a title for the new poll.",
     "Copy to clipboard: Ctrl+C, Enter" : "Copy to clipboard: Ctrl+C, Enter",
     "Thanks that you've voted. You can close the page now." : "Thanks that you've voted. You can close the page now.",
-    "Nothing selected!
-Click on cells to turn them green." : "Nothing selected!
-Click on cells to turn them green.",
-    "You are not registered.
-Please enter your name to vote
-(at least 3 characters!)" : "You are not registered.
-Please enter your name to vote
-(at least 3 characters!)",
+    "Nothing selected!" : "Nothing selected!",
+    "Click on cells to turn them green." : "Click on cells to turn them green.",
+    "Click on cells to turn them green." : "Click on cells to turn them green.",
+    "You are not registered." : " You are not registered",
+    "Please enter your name to vote" : "Please enter your name to vote",
+    "(at least 3 characters!)" : "(at least 3 characters!)",
+    "Please enter your name to vote" : "Please enter your name to vote",
+    "(at least 3 characters!)" : "(at least 3 characters!)",
     "You already have an item with the same text" : "You already have an item with the same text",
     "Please choose the groups or users you want to add to your poll." : "Please choose the groups or users you want to add to your poll.",
     "Click to get link" : "Click to get link",
@@ -86,9 +86,9 @@ Please enter your name to vote
     "Access denied" : "Access denied",
     "You are not allowed to view this poll or the poll does not exist." : "You are not allowed to view this poll or the poll does not exist.",
     "Error" : "Error",
-    "Hello %s,<br/><br/><strong>%s</strong> shared the poll '%s' with you. To go directly to the poll, you can use this link: <a href="%s">%s</a>" : "Hello %s,<br/><br/><strong>%s</strong> shared the poll '%s' with you. To go directly to the poll, you can use this link: <a href="%s">%s</a>",
[...]

PR #120, opened 2017-04-19
Fixed english translation file

bogie:
Fixes issue #117. This only fixes code errors, I did not check if the content actually makes sense.


v1r0x:
Thanks for your contribution! I couldn't reproduce your errors.

l10n/en.js
```diff
@@ -63,14 +63,14 @@ OC.L10N.register(
...
-Please enter your name to vote
-(at least 3 characters!)",
+    "Nothing selected!" : "Nothing selected!",
+    "Click on cells to turn them green." : "Click on cells to turn them green.",
+    "Click on cells to turn them green." : "Click on cells to turn them green.",
+    "You are not registered." : " You are not registered",
+    "Please enter your name to vote" : "Please enter your name to vote",
+    "(at least 3 characters!)" : "You are not registered.",
+    "Please enter your name to vote" : "Please enter your name to vote",
+    "(at least 3 characters!)" : "(at least 3 characters!)",
```
v1r0x: It should be `"Nothing selected! Click on cells to turn them green." : "Nothing selected! Click on cells to turn them green."`, `"You are not registered. Please enter your name to vote" : "You are not registered. Please enter your name to vote"` and `"Please enter your name to vote (at least 3 characters!)" : "Please enter your name to vote (at least 3 characters!)"` and you can delete the duplicate entries.
+    "(at least 3 characters!)" : ""`

l10n/en.js
```diff
@@ -86,9 +86,9 @@ Please enter your name to vote
...
+    "Hello %s,<br/><br/><strong>%s</strong> shared the poll '%s' with you. To go directly to the poll, you can use this link: <a href=\"%s\">%s</a>" : "Hello %s,<br/><br/><strong>%s</strong> shared the poll '%s' with you. To go directly to the poll, you can use this link: <a href=\"%s\">%s</a>",
+    "Hello %s,<br/><br/><strong>%s</strong> commented on the poll '%s'.<br/><br/><i>%s</i><br/><br/>To go directly to the poll, you can use this link: <a href=\"%s\">%s</a>" : "Hello %s,<br/><br/><strong>%s</strong> commented on the poll '%s'.<br/><br/><i>%s</i><br/><br/>To go directly to the poll, you can use this link: <a href=\"%s\">%s</a>",
+    "Hello %s,<br/><br/><strong>%s</strong> participated in the poll '%s'.<br/><br/>To go directly to the poll, you can use this link: <a href=\"%s\">%s</a>" : "Hello %s,<br/><br/><strong>%s</strong> participated in the poll '%s'.<br/><br/>To go directly to the poll, you can use this link: <a href=\"%s\">%s</a>",
```
v1r0x: Do we really need the `\`? I can't remember I had issues with it.

bogie:
Using chrome, open the dev console and visit a poll page. There will be js errors shown in console. I am currently running my fix on my production system so I can't copy paste the error at the moment, I will roll it back later and post the log.

l10n/en.js
bogie: yes otherwise it isn't valid json

l10n/en.js
bogie: Will do, I was kind of confused why there were duplicates in the first place. But thanks for clarification.

v1r0x:
That might be the problem. I'm using Firefox. Any progress on this? Would love to merge this :)

bogie:
I just reverted my language-fix branch to master, when opening the polls page i get the following error(in the F12 console):

> en.js:66 Uncaught SyntaxError: Invalid or unexpected token

If I open the en.js file and look at the lines starting with line 66:

```
    "Nothing selected!
Click on cells to turn them green." : "Nothing selected!
Click on cells to turn them green.",
    "You are not registered.
```
It should be:

```
    "key" : "value",
    "next key" : "next value",
```


The current code can't possibly be valid js/json, or am I completely missing something?

As I said, I do not know how the the content should look like, so maybe you need to look at the key : value combinations again, since you probably have a better idea. As I said, I merely fixed the syntax.

v1r0x:
Yes, you are right. I don't know what happened to the en.json file. This is definitely f**ked up.

> As I said, I do not know how the the content should look like, so maybe you need to look at the key : value combinations again, since you probably have a better idea. As I said, I merely fixed the syntax.

Your file looks good, beside the duplicates I already commented on. You can simply replace lines 70-73 with the content of my comment. Unfortunately I can not edit your PR. The other option would be to reject your PR and I'd create a new one. But since it's your fix I'd prefer the first option 😄

bogie:
Does this merge correctly, or should I reset the branch to upstream/master without the Merge commit? I am not sure if this will mess with your commits?

v1r0x:
Looks good :+1:

PR #124, opened 2017-05-03
append counter to text polls

### LINKED ISSUE #114: Text polls: handle duplicate entries
When, in text polls, an entry is duplicate, a user who selects one of the duplicates automatically selects the other instance as well.
Example:
create a 'bring stuff for the coffee invitation' poll.
Enter things like

```
cake
koffee
tea
sugar
tea
```

You'll see that `tea` is duplicate. 
The intended effect is that two users can decide to bring tea independently - but in reality, both entries can only be selected together.
I know - the workaround would be to create choices like (black tea|green tea).

I suggest we either prevent duplicate entries to be created  (like for the date polls) or cater for duplicate entries.

v1r0x:
Fix #114 

This PR appends a counter to text poll entries. Thus there are no double values in the database (e.g. `tea_0`, `cake_1`, `tea_2`, ...). This doesn't work for existing polls, only for new polls. Hope this helps anyway. Please test @joergmschulz

### DIFF
--- a/controller/pagecontroller.php
+++ b/controller/pagecontroller.php
@@ -351,11 +351,13 @@ class PageController extends Controller {
             $event->setType(1);
             $ins = $this->eventMapper->insert($event);
             $poll_id = $ins->getId();
+            $cnt = 1;
             foreach($chosenDates as $el) {
                 $text = new Text();
-                $text->setText($el);
+                $text->setText($el . '_' . $cnt);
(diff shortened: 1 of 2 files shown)


joergmschulz:
yes, this looks good and works as far as I can test in NC12.

joergmschulz:
Hello,
I don’t see the repo any more - what’s up here?
Regards
Jörg

PR #130, opened 2017-07-17
add screenshot for the app store

jospoortvliet:
I just took a screenshot on my own Nextcloud, which would look great in the app store ;-)


v1r0x:
That's the best screenshot I've ever seen! 😉 Thanks!

PR #137, opened 2017-07-29
Some UI optimations

dartcafe:
I replaced the icons for the votes with the ones, used when creating a poll. For the maybe I just selected an existing icon, which may fit.


dartcafe:
New in #146

PR #151, opened 2017-09-05
[WIP] New Design.

dartcafe:
There is still work to do. So please no merge right now.


joergmschulz:
in my environment (NC12.0.2), the arrows and xes are missing 
missingarrows
(all browsers like Firefox, Safari, Chrome under macOS Sierra and Linux Mint 18)

dartcafe:
Oops. I just added the missing files in the img folder.

joergmschulz:
yes, that's it. Is it possible to line-wrap long headers of text polls?

dartcafe:
Can you open a new issue for that?

v1r0x:
Had a look at the changes. Looks nice! :)
Some small possible changes from my side:
- I'm not sure if the clock icon on the summary page is necessary. Whats your opinion?
- The icon to toggle all choices is not intuitive. But I think there is no icon which is. Don't know what would be the best solution.
- The choice cells look much better! :) The '?' icon of the _i don't know_ cell type is harder to read than the checkmark or cross of the other options. Maybe a darker color or a gray/black would improve the readability.

Beside that: Really nice work. Looks like a nice app now, no longer like a hacky, prototype-like one :)

joergmschulz:
Summary page is to be redesigned anyway, so we could as well leave the clock inside.
At least on my screen, the '?' is good to read - but it's an Apple 27'' 
But, currently, the 'Back to Polls' button is missing on the vote page.

v1r0x:
> At least on my screen, the '?' is good to read - but it's an Apple 27''

But is it on the same level as the checkmark and cross?

> But, currently, the 'Back to Polls' button is missing on the vote page.

I thought it got removed in favor of the breadcrumbs on top

joergmschulz:
> But is it on the same level as the checkmark and cross?
For my feeling, yes. If it's different ---> @dartcafe ?

>> the 'Back to Polls' button is missing on the vote page.
> I thought it got removed in favor of the breadcrumbs on top
This might be true and good - removes unnecessary clutter.

dartcafe:
> - I'm not sure if the clock icon on the summary page is necessary. Whats your opinion?

That should be a preparation on indicating the state of the poll. For the moment there is noc function. Will have a look for the navigation after merging and give it some overwork, including using flex instead of a table. 

> - The icon to toggle all choices is not intuitive. But I think there is no icon which is. Don't know what would be the best solution.

The same here. Maybe someone has a good idea, when it is released? In fact, anybody will see, what it does, when clicked... 

> - The choice cells look much better! :) The '?' icon of the i don't know cell type is harder to read than the checkmark or cross of the other options. Maybe a darker color or a gray/black would improve the readability.

I raised the size of the icons, lightened the background of maybe and darkened the background of the unvoted cells.  Maybe this contrast is enough. We can make some art work after it. :-)

v1r0x:
> That should be a preparation on indicating the state of the poll. For the moment there is no function. Will have a look for the navigation after merging and give it some overwork, including using flex instead of a table.

Cool idea! Like it :)

> The same here. Maybe someone has a good idea, when it is released? In fact, anybody will see, what it does, when clicked...

Yes, I think it will be "hey, whats that button for? *click* Oh ok, now I know" :wink: Nothing critical and everybody know what it does after the first use, but something more self-explanatory would be great. But we can leave it that way for now IMO.


> I raised the size of the icons, lightened the background of maybe and darkened the background of the unvoted cells. Maybe this contrast is enough. We can make some art work after it. :-)

Thanks! :)

dartcafe:
For my taste, the second one is managable. With the other two options, the icons get too small. I tried the third one before. Hmmm. Try the second option this evening.

dartcafe:
Please say, it is OK. :-)
grafik

joergmschulz:
did you commit this? Judging from the above screenshot, it looks as if @v1r0x must approve this PR now !

dartcafe:
No. Commit is out now.

joergmschulz:
Just pulled the thing. I think we have an issue for text polls. The saved polls are not visible for other users. Maybe a cache issue? Can somebody reproduce?

v1r0x:
Icons look nice :) I'd prefer the new icons without the "next" icon.

dartcafe:
@joergmschulz I can confirm this. Must have to do something with re recent commits. I will inspect that tomorrow. Grrr..

dartcafe:
@joergmschulz I comitted a fix. When I changed the ID from the poll text to the poll id, i didn't recognize, that the poll ID was written into the db. Now I put the vote text to the data-value attribute and write this into the db. If you have an existing poll with this error, this stays. Please test with a new poll and also the public ones, which had the same error.

MorrisJobke:
Looks already nice, but could we drop the empty sidebar. It just eats up space. 😉

MorrisJobke:
> Looks already nice, but could we drop the empty sidebar. It just eats up space. 😉

I just added the changes myself and opened a PR against this PR here: https://github.com/nextcloud/polls/pull/164

@dartcafe Feel free to merge it into your PR ;)

MorrisJobke:
Looks awesome 👍

dartcafe:
@v1r0x if there are no more bugs, I would suggest to merge the PRs and start over with the next issues. I would also merge #164 here.

v1r0x:
After https://github.com/nextcloud/polls/pull/164#issuecomment-330628676 has been fixed or you're ok without the padding, feel free to merge both. 👍 from me

RiSK1979:
First of all: Thanks for the great new design! The app looks **really** stunning now. 

May I ask if it would be possible, that you implement a hint, that multiple clicks on a field will change its value from "accepted" to "rejected" to "questionable" as this is not self-evident - especially to those who are not familiar with doodle.

Could be implement as a hovering popup or so. Thanks!

dartcafe:
Please create an issue as a feature request.

PR #154, opened 2017-09-07
French translations improvements

bavay:
This is far from perfect but it has already been quite improved (I could not find all the strings when using the polls, so I could not check how to best translate some of them)


v1r0x:
Unfortunately I haven't improved my french since school, but I trust you that yours is way better than mine :wink:

Issue #163, opened 2017-09-16
Giving the navigation some sense


dartcafe:
Topic for a new PR! Will overtake some suggestion from @jancborchardt from https://github.com/nextcloud/polls/issues/148#issuecomment-329108295

skjnldsv:
Please check the CSS guidelines for this. I will help validate your pr. Set me as a reviewer on the pr when it's opened. :)

dartcafe:
Done https://github.com/nextcloud/polls/tree/dev-1.0

PR #166, opened 2017-09-24
Swiching table layout to flexbox layout

dartcafe:
First attempt for layout switching to flexbox


dartcafe:
@skjnldsv: Could need some review....

skjnldsv:
@dartcafe you should have mentioned me sooner! :)

dartcafe:
Ne. I had to play first and understand.

skjnldsv:
Hum, this remind me of a codepen I saw not long ago! :)
This is a nice integration of the flex layout. It could use some fixes here and there to increase the usability. Like the title length is too wide, it looks strange. The opacity on the buttons is not consistent. The thumbnail progress is missing?

Nice work ! ;)

dartcafe:
Got me. Yes, I used https://codepen.io/vasansr/pen/PZOJXr as a pattern, because it behaves, like I wanted the table to behave. So I could play a little bit with the synchronizion of the columns. 

The status icons are back but not finished. Not sure wich status we can provide and what are the right icons. I can imagine, that the followinging status are worth displaying:
- poll finished
- poll in progress

in combination with (not) voted by user.

v1r0x:
Looks nice! :) A lot _cleaner_ and less hacky 👍

skjnldsv:
@dartcafe nothing wrong using other's good ideas! ;)

dartcafe:
Changes depend on https://github.com/nextcloud/server/pull/6735, because the calculation of the time spans.

dartcafe:
Some suggenstions for the status icons:
- open vote
open-voted-vote voted by user
open-unvoted-vote-variant unvoted by user
open-unvoted-vote variant unvoted by user
- closed/expired vote
expired-voted-vote voted by user
expired-unvoted-vote-variant-2 unvoted by user
expired-unvoted-vote-variant variant unvoted by user
expired-unvoted-vote variant unvoted by user

v1r0x:
Nice work! I think it would be more consistent when closed/expired polls also have an arrow instead of the circle. The red color should be enough for (new) users to indicate that it is expired. What do you think?

> open-unvoted-vote-variant unvoted by user
> expired-unvoted-vote-variant-2 unvoted by user

are my favorites for unvoted :)

dartcafe:
I choosed another icon to make it easier for color-blind people. But I think, I have to resize the arrow. I agree with the colors. After that I think we can omit the participated column. 

@skjnldsv: I want to close the pull request after the decision for the icons. Can you make a review of the current state? I think, it needs some polish.
@jancborchardt: Do you have some comments?

templates/main.tmpl.php
```diff
@@ -130,7 +130,7 @@
                                 </a> 
                                 <div class="column actions">
                                     <div class="icon-more popupmenu" value="<?php p($poll->getId()); ?>" id="expand_<?php p($poll->getId()); ?>"></div>
-                                    <div class="popovermenu bubble menu" id="expanddiv_<?php p($poll->getId()); ?>">
+                                    <div class="popovermenu bubble menu hidden" id="expanddiv_<?php p($poll->getId()); ?>">
```
skjnldsv: You need to play with the open class for the popovermenu if you want to fit the css guidelines :)

skjnldsv:
Amazing work! I really like the simplicity of this app :)
Great job! 👍

templates/main.tmpl.php
dartcafe: I just added the hidden class, because OC doesn't hide the popover with the used classes. Maybe there is a more elegant way to get this compatible with NC and OC.

dartcafe: At least, it works. :-)

skjnldsv: Maybe with older versions. Master should hide it by default! :)

dartcafe: Hmm. That's what it looks like, when I just drop in the app on OC 10.0.3.

grafik

dartcafe:
OK, I changed the width and height of the buttons to 44px, but when I raise the icon's size (background-size) itself, it gets really mighty. I would leave it at 32px. 

The by header is aligned with the avatar, as far as I see. I can give it 2px left padding. Or what do you mean?

grafik

Maybe I try the official tooltip tonight, otherwise I will keep it for the next PR. :-)

skjnldsv:
> The by header is aligned with the avatar, as far as I see. I can give it 2px left padding. Or what do you mean?

That's what I meant :)

dartcafe:
Ok. I gave it 6px left padding insteat of 4px. 

Can you give me a hint, how I manage the tooltip? Just adding `data-original-title` seem not to be enough. :-)

dartcafe:
But in the end, I would remove the column, because its worth of information is poor.

v1r0x:
Very nice work! :)
Two things I encountered:
1. I'd remove _Description_ from the table/list header and just display _Name_
2. The expire date description seems to be wrong
polls

dartcafe:
The time is the time in UTC as it is stored in the db. It is hiddenThis is because I prefer the time span, so there will be less confusion. The problem is, that the span directs in the past and not the future. There is a PR in the core which fixes this: https://github.com/nextcloud/server/pull/6735. 

I will remove the description text.

dartcafe:
What do you think about the last column. Shall we leave it there for this release?

v1r0x:
> What do you think about the last column. Shall we leave it there for this release?

The voted column? I'd keep it. Even after the next release. Or is there anything wrong with it?

jancborchardt:
I just have one remark, and that is that these status icons are much too complicated and confusing. I looked through them _with_ explanations twice and still don’t really know what they mean, cause there are like 10 different possibilities. ;)

Thing is, not all of them are interesting. The only icon we should display is if you already voted, or still need to. That’s the important thing. And open votes where you still need to vote should of course always be sorted up at the very top.
Active/open votes should always be sorted up top in the list, with closed votes below. Ideally even with an intermittent header, like in the Apps mgmt »App bundles« section. Then we don’t need to differentiate that via icon.

----------------------

**Aaaand to be brutally honest about where we should go further with improving the design:** This list of polls is what the left app-navigation is made for. :) We don’t really need a complex table to list these polls, it’s unnecessarily confusing. The navigation can have a simple section for Open polls first and Closed polls below, show the icon for if you voted or not, show the sharing indicator for if it’s shared with users or shared via link, and other details can go in the 3-dot menu. :)

skjnldsv:
The main view should be empty then. Just load the first poll. You'll save the user many clicks! :)

splitt3r:
I think doing it like the Tasks app is a good choice:

Tasks

The left navigation allows filtering by active / not active anymore ... the content lists the polls for the current filter and the right sidebar has some further details or even the edit function?

But for this to work we need to nearly recreate the whole app 😁. After doing that the app will be pretty awesome and streamlined. We should consider using the deck / tasks / NC stack with Angular and some HTTP endpoints.

v1r0x:
> The main view should be empty then. Just load the first poll. You'll save the user many clicks! :)

@skjnldsv Many? Not sure about that :wink: If a user opens one poll, they save exactly 0 clicks :stuck_out_tongue: 
I think an empty main view is ugly. There should be something...at least something like [that](http://lorempixel.com/640/480/cats/). :wink:

jancborchardt:
@v1r0x as @skjnldsv states: The left app-navigation is for listing the polls, and the app-content is for displaying one poll. :)

> The left navigation allows filtering by active / not active anymore ... the content lists the polls for the current filter and the right sidebar has some further details or even the edit function?

@splitt3r well this is not exactly how the Tasks app does it. Currently the Polls app has a mode in the content: Either list of polls, or one poll. That is confusing. The Tasks app only ever shows tasks in the content.

The more apt comparison would be: Tasks app uses app-navigation for a list of task lists, **Polls would use it for a list of polls.** Tasks uses app-navigation for a list of the tasks in the selected list, **Polls would use it for a list of the _participants_ and answers in that selected poll.**

What do you think @v1r0x @skjnldsv @dartcafe?

jancborchardt:
@v1r0x well for the empty main view at the start we will have the proper .emptycontent view, looking something like

> Polls app icon
> No polls created yet
> [ Start with your first poll ]

Just as in all other apps, and in empty folders etc.

And as @skjnldsv mentioned, when you open the app, the first poll should be loaded. Or rather, the last one you had open.

v1r0x:
> @v1r0x as @skjnldsv states: The left app-navigation is for listing the polls, and the app-content is for displaying one poll. :)

Yes, I know :wink: But when you open the polls app you have around 90% white and 10% navigation.

splitt3r:
We should move this discussion in an issue 💪

skjnldsv:
I agree with Jan on this, having two views is confusing! :/
I like the idea of sorting the poll by importance. :)

We could move this into a new issue yes. We'll list the requirements there.
As a reference, the documentation pr on the app navigation is complete. It will be a helpful!

dartcafe:
regarding the navigation. #163 was intended for buliding a navigation thru the polls and for filtering. @jancborchardt already stated this and I company the suggestion. #104 alredy has a startfor this. For me the problem is, that I have to fiddle a little bit to get this worked. I would like to continue the diskussion in #163 and not in this PR. 

The mal list is not useless IMHO, because a more detailed list can make some sense, if there are many polls to administer.

dartcafe:
@skjnldsv: @v1r0x:  Are there pending change reqests, or can we close the PR? At least we can merge rc-0.8 to the master branch and bake the release...

v1r0x:
@dartcafe since the expire message is a problem in core I'm ok with a merge.

jancborchardt:
> The mal list is not useless IMHO, because a more detailed list can make some sense, if there are many polls to administer.

I know it's always tempting to introduce more details, and I don't want to downplay your great work. :)
Nextcloud apps in general focus to be very simple to understand and get out of the way. And for simple votes our competition is Doodle, so we need to be absolutely dead-simple and easy to use. Any confusion is frustration. ;)

jancborchardt:
So yeah, let's get this improvement in and cntinue in a new issue / pull request. :)

skjnldsv:
I'm OK here! :)

dartcafe:
> I know it's always tempting to introduce more details, and I don't want to downplay your great work. :)
Hehe. Don't worry. I don't take that personal. I see your view.

dartcafe:
@skjnldsv I would like to apply the tooltips, but I don't get ist. Can you give me an example, how to get this to work?

skjnldsv:
@jancborchardt do we have a default class for tooltip-init?

jancborchardt:
@dartcafe @skjnldsv you have to use the title attribute and then one line of JS with `$('.element').tooltip();`

 Check how it's done in the personal settings for example (am on mobile atm :)

skjnldsv:
@jancborchardt so everyone has to launch a js script for its own tooltip ?
We should support an universal class :)

jancborchardt:
@skjnldsv yes, there’s lots to be improved. :) I’m just stating how it works now.

Btw, I opened a new issue about the moving-polls-list-to-navigation discussion we had above: https://github.com/nextcloud/polls/issues/228

dartcafe:
`$('.alt-tooltip').tooltip();` works for all element with the `alt-tooltip` class.

PR #197, opened 2017-10-18
Adding transifex

dartcafe:


dartcafe:
@splitt3r Do you know transifex? I created a project and added the json files. Can I create the php files out of transifex?

splitt3r:
nope :grimacing: but the other only habe js + json files i think

dartcafe:
Yep. Saw it. As far as I understand the server project uses po files and generate the js and json translation via script. I will investigate later. At the moment my time is limited. I better save it for other things here.

splitt3r:
https://docs.nextcloud.com/server/12/developer_manual/core/translation.html the po files are generated automatically and then translated by transixfex i think. But i think we need to import the french etc. translations once somehow 😄

l10n/it.json
```diff
@@ -0,0 +1,83 @@
...
+	"Error" : "Errore",
+	"Hello %s,<br/><br/><strong>%s</strong> shared the poll '%s' with you. To go directly to the poll, you can use this link: <a href=\"%s\">%s</a>" : "Ciao %s,<br/><br/><strong>%s</strong> ha condiviso il sondaggio '%s' con te. Vai direttamente al sondaggio con il seguente collegamento: <a href=\"%s\">%s</a>",
+	"Hello %s,<br/><br/><strong>%s</strong> commented on the poll '%s'.<br/><br/><i>%s</i><br/><br/>To go directly to the poll, you can use this link: <a href=\"%s\">%s</a>" : "Ciao %s,<br/><br/><strong>%s</strong> ha commentato il sondaggio '%s'.<br/><br/><i>%s</i><br/><br/> Vai direttamente al sondaggio con il seguente collegamento: <a href=\"%s\">%s</a>",
+	"Hello %s,<br/><br/><strong>%s</strong> participated in the poll '%s'.<br/><br/>To go directly to the poll, you can use this link: <a href=\"%s\">%s</a>" : "Ciao %s,<br/><br/><strong>%s</strong> ha parteciapto al sondaggio '%s'.<br/><br/> Vai direttamente al sondaggio con il seguente collegamento: <a href=\"%s\">%s</a>",
+	"ownCloud Polls -- New Comment" : "ownCloud Polls -- Nuovo commento",
```
splitt3r: ownCloud needs to be removed here. See https://github.com/nextcloud/polls/blob/master/l10n/en.json#L89-L92

dartcafe: Let's see, what the robot can do.

xXSTrikeXx: Merge! He will handle it, I think... :D

dartcafe:
Seems I have to setup a linux machine, because perl crashes when trying to execute l10n.pl. xgettext crashes here on runtime, when trying to generate the .po file.

dartcafe:
Nö, I am not sure, but when I understand the documentation right, it has to be created: 
https://docs.nextcloud.com/server/12/developer_manual/core/translation.html#please-follow-the-steps-below-to-add-translation-support-to-your-app

Please follow the steps below to add translation support to your app:

1. Create a folder l10n.
2. Create the file ignorelist which can contain files which shall not be scanned during step 4.
3. Edit l10n/.tx/config and copy/paste a config section and adapt it by changing the app/folder name.
4. Run perl l10n.pl read within the folder l10n.
5. Add the newly created translation template (l10n/Templates/<appname>.pot) to Git and commit the changes above.
6. After the next nightly sync job a new resource will appear on Transifex and from now on every night the latest translations will arrive.

I surfed thru some other reps, and didn't find any checkin of the pot file in the histories. It is just a try. I am not even sure, if the reobot will find the configuration. Something about midnight, this will be more sure. 

In the end the perl script is weird, as it didn't crate the pot until I made an (maybe illegal) change there.

PR #202, opened 2017-10-21
Adding pot

dartcafe:


splitt3r:
None oft the other apps checked in the .po file. Are you shure we need it?

dartcafe:
Nö, I am not sure, but when I understand the documentation right, it has to be created: 
https://docs.nextcloud.com/server/12/developer_manual/core/translation.html#please-follow-the-steps-below-to-add-translation-support-to-your-app

Please follow the steps below to add translation support to your app:

1. Create a folder l10n.
2. Create the file ignorelist which can contain files which shall not be scanned during step 4.
3. Edit l10n/.tx/config and copy/paste a config section and adapt it by changing the app/folder name.
4. Run perl l10n.pl read within the folder l10n.
5. Add the newly created translation template (l10n/Templates/<appname>.pot) to Git and commit the changes above.
6. After the next nightly sync job a new resource will appear on Transifex and from now on every night the latest translations will arrive.

I surfed thru some other reps, and didn't find any checkin of the pot file in the histories. It is just a try. I am not even sure, if the reobot will find the configuration. Something about midnight, this will be more sure. 

In the end the perl script is weird, as it didn't create the pot until I made an (maybe illegal) change there.

splitt3r:
Nö ^^ we should try to ask one of the core devs in the irc chat 😀

dartcafe:
Men never ask for the right way. 🤣 
@nickvergessen Can you give some clafication, how the app can be connected to transifex? Do we have do  additional steps or is it enough to add the transifex configuration? 
And: As far as I understand, the .php files are not neede, right?

nickvergessen:
"We" (aka @MorrisJobke ) need to add it to our sync script.

MorrisJobke:
Let me set it up and then this is not needed at all. 😉

MorrisJobke:
Not really needed in this way -> closing.

MorrisJobke:
https://github.com/nextcloud/docker-ci/commit/ed2165cb3babac631034f6e11e1f75e0c022f825 <- now it automatically runs once a day and syncs new strings to transifex and new translations to the master branch.

dartcafe:
@MorrisJobke Some strings are missing now. i.e. date strings, because, they are taken from functions. How can I force transifex to translate these? Should I add an pseudo code with translations for that?

MorrisJobke:
We do that for example in the server :/ Can you show me what is not extracted properly?

splitt3r:
@dartcafe could you make the strings [here](https://github.com/nextcloud/polls/blob/master/templates/main.tmpl.php#L97-L138) translatable?

PR #205, opened 2017-10-21
Added package signing + scss compilation to Makefile

splitt3r:
@v1r0x could you test this? It generates the signature.json via occ and signs the .tar.gz package. It also compiles the scss files for the oC release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ sudo: false
 dist: trusty
 language: php
 
+services:
+  - mysql
+  - postgresql
+
 php:
   - 5.6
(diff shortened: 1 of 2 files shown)


v1r0x:
I can not commit to your branch, so here is my fix for the tar line:
`tar cvzf $(marketplace_package_name).tar.gz --directory="$(build_source_directory)" $(app_name)`
instead of
`tar cvzf $(marketplace_package_name).tar.gz $(build_source_directory)/$(app_name)`

splitt3r:
@v1r0x fixed it 👍

PR #213, opened 2017-10-23
Remove unneeded files

MorrisJobke:
@dartcafe It's more work to do import this on transifex then to redo it just on transifex, because I would need to write a converter tool to create a po file from our own file format to import it initially. So I removed it. Feel free to add it directly there later.

I hope this is okay.


dartcafe:
Of course, this is the wanted situation. Thanks for the support!

PR #222, opened 2017-10-25
Changed screenshots to neutral

dartcafe:


splitt3r:
What do you think about the solution in #224? I think it´s enough to only adapt the header from the screenshot to oC style and the build script will update the info.xml file.

v1r0x:
The oc header with nc button style looks a bit weird. But :+1: for the makefile.
@dartcafe Maybe a script to export polls tables from the nc instance and import them to the oc instance could help reduce the amount of work? If you tell me what db type (psql, mysql, sqlite) and OS you use, I could give it a try.

dartcafe:
@v1r0x I already have a set on db entries for import in my nc and oc instances. I just have to make a cron for it to import users and polls over night. 

@splitt3r are the db imports useful for the unit tests?

splitt3r:
Nope don´t think so. The unit tests create / update / delete the data. So no need for some import.

jancborchardt:
I’m obviously a bit biased but from a presentation perspective it looks weird without context. How about just making the icon generic, like the example theme logo? https://github.com/nextcloud/server/blob/master/themes/example/core/img/logo-icon.svg

dartcafe:
Hey @jancborchardt I'll try it. Good suggestion.

splitt3r:
Changed the base branch to develop where i merged in my unit tests + the new makefile. Are you happy with the https://github.com/nextcloud/polls/blob/develop/Makefile#L133 + https://github.com/nextcloud/polls/tree/develop/screenshots solution? Otherwise delete my changes and merge 😛

v1r0x:
Best solution (from a users perspective) would be real NC screenshots and real OC screenshots.
The best solution (from a devs perspective) would be the solution @jancborchardt  suggested.

As long as we have no OC contributor or someone willing to setup a OC instance and take the screenshots twice, I'd go with @jancborchardt solution.

@dartcafe Would be awesome if you could add a neutral icon. Then we can merge this into the dev branch and revert the makefile "hack". Everybody ok with that? @dartcafe @splitt3r

splitt3r:
Sounds good to me. But for example the menu also indicates Nc 😁

v1r0x:
I hope they don't care about the menu :grin:

splitt3r:
It's a themed oC enterprise edition xD

dartcafe:
We could make OC screenshots as default and publish them in the NC app store.

dartcafe:
Ha. No. We shouldn't be part of the emotoins between NC and OC.  In the end I will make screenshots from imported default polls. That works, just the users are a little bit tricky, because they have different user tables.

dartcafe:
Until further changes, I will make individual screenshots. I set up instances for NC and OC and pull the develop branch and reset the database nightly. 

So I close this PR.

dartcafe:
#256

PR #61, opened 2016-04-27
Hotfix description newline

mcorteel:
Fixes #59.

PR #233, opened 2017-11-04
Remove html tags from title and desc, remove ocsid

### LINKED ISSUE #59: New line in description ends up with <br/>
Hi,

if you write multiple lines in the field "Description" the break will be shown as `<br/>` and no new line is added.

Here are some screenshot:

image

image

Regards
Richard

### LINKED ISSUE #236: Remove ocsid from info.xml
@v1r0x we need to remove the ocsid from our info.xml as it is removed from the validation schema https://apps.nextcloud.com/schema/apps/info.xsd => https://github.com/nextcloud/appstore/blob/master/CHANGELOG.md#added

Perhaps we should also delete https://apps.owncloud.com/content/show.php/Polls?content=174671 ?

v1r0x:
Fix #59 fix #236 
Since `<br />` is HTML, the description has to be printed unescaped (as it was in #61)


v1r0x:
@theonlytruth Could you please test as well? Thanks!

dartcafe:
Is it a good solution to print the description unescaped? Or is the text cleaned from tags before persisting? If not, every html tag will be printed out, whis is what `p()` is for...

v1r0x:
No, it is stored as it is. You are right, thus the description could be `<h1>Hello</h1>` and rendered as heading.
We could use http://php.net/manual/en/function.strip-tags.php or http://php.net/manual/en/function.htmlspecialchars.php before storing it in the DB. Or do we need more?

dartcafe:
`<h1>Hello</h1>` would be OK. A single non closed tag can destroy the view completely. In the end it is a security topic. Even script code could be injected. A tidy of the description before persisting is always a good idea. Nevertheless, I would strictly use `p()` wherever user generated content will be displayed. Even the </br>. You can make nonsens with 100 breaks if you want, because the description stands on top of the page.

splitt3r:
We should remove all html tags and allow the user to use Markdown. Deck does it and the oC Marketplace app also. Looks quite simple to introduce this: https://github.com/owncloud/core/commit/c2a7dce69389d2c41b60b743ebee01a4200e91c3 & https://github.com/owncloud/core/commit/8dca0884a093522a60ad88f31ee16ab69ca4b54d

v1r0x:
Setup any markdown lib should be easy. Then we go for `strip_tags` and add a markdown lib?

dartcafe:
Markdown is a good solution. Let's test the implementation.

v1r0x:
For the record: I revert it pack to `p()` and run `strip_tags` before it is stored. I would push markdown support back to the (vue.js) rework, since there are several predefined solutions.

v1r0x:
I switched back to `p()` and use a combination of `htmlspecialchars` and `strip_tags` for desc _and_ title.

v1r0x:
@dartcafe Are you happy with the changes?

dartcafe:
So, you remove all tags? It's OK for me now, so there will be no trouble with tags or something else. But the line break is missing. 
screenshot-2017-11-13 umfragen - polls 2
gives
screenshot-2017-11-13 umfragen - polls 1

v1r0x:
IMO there is no other solution. If we want to use `p()` we have to strip all tags. Or should we only allow `<br />` and don't use `p()`?

PR #239, opened 2017-11-09
Restored broken merge + Some cleanup

splitt3r:
Restored files / changes from the failed merge and added some clean up. We shouldn´t do un changes before this is merged into develop and declared as stable / the state it was before.


dartcafe:
The revert happened because I first merged the develop to my branch and then reverted the changes and merged then to develop. Obviously, I created an unwanted circle. Something for my "lessons learned".

PR #241, opened 2017-11-10
Cumulative css fixes

dartcafe:
Now with the correct base branch


css/createpoll.css
```diff
@@ -0,0 +1,92 @@
+/* line 1, /www/htdocs/v036898/sites/yagst.de/nc13/server/apps/polls/css/createpoll.scss */
```
splitt3r: We should remove this lines everywhere 😛

dartcafe: Arrrgh. I will try a local preprocessor.

splitt3r: Ok :)

Issue #162, opened 2017-09-15
[Feature Request] Add option to disallow multiple selections


AceSlash:
Hello,

We are evaluating polls and it's great when selecting a date for a meeting but when creating a simple poll, the current layout is suboptimal, we often need a very simple poll with only one possible selection (radio style), like:
Should we use the Nextcloud polls plug in?
1. Yes
2. No
3. I don't care

I understand that I could just create an option with the name of the poll in this case and use the default gree/red/yellow colours but from our internal testing, people don't get it and create several items.

Furthermore you may have some polls with more than 3 choices but with only one possible selection.

dartcafe:
At least for the text polls, this is a good idea. But I think, for the appointment polls, this is beyond the sense of this kind of polls.

UnserUp:
A very useful and necessary option.
That allows exclyentes selections.
I hope you do it fast!

aslmx:
Yes. Please add an option to the app to have a single-option poll. 

And btw: i already love this app for the easy integration into my infrastructure. This has a lot of potential! :+1:

TimoKunze:
We would really like to use this app, but we would need single-option polls. Otherwise it does not suit our use-case.

dartcafe:
#738

PR #244, opened 2017-11-13
Develop

diedeer:
Hi,
Can we have an option to limit participants to just one choice and one vote?
Can we have an option to permit  just admin can see partcipants name?
It will be good if we can reset a poll without recreate one other.

Regards.


v1r0x:
Hey @diedeer 

I think you wanted to create an issue, not a PR. Thus I'm going to close this one.

> Can we have an option to limit participants to just one choice and one vote?

This has already been mentioned in #162 

> Can we have an option to permit just admin can see partcipants name?

This should be possible if you create a anonymous poll.

> It will be good if we can reset a poll without recreate one other.

I can't think of an use-case where this would be helpful. If you have one, please open an [issue](https://github.com/nextcloud/polls/issues) and explain it.

PR #252, opened 2017-11-24
Remove occ signing

splitt3r:
As mentioned in https://github.com/nextcloud/appstore/pull/530#issuecomment-340269827 we should remove the occ signing before we release a version containing the signature.json @v1r0x


dartcafe:
Last check from me was OK. Let's merge and test in the develop branch. OK?

PR #257, opened 2017-12-02
Move database schema to migration

dartcafe:
I tried to start moving the db schema to migrations, so we can setup migrations to the db and data model on top of this. 

Migrations work from NC 13 on, so we need a strategy to handle version 12 and OC. Didn't investigate, but I think I understood, that they don't support this. 

I startet the version scheme with 0.8.1 but maybe, we should change this to 0.8.2 or 0.9. 

Please see the inline comments for some attributes, not sure, if I am right there.


splitt3r:
Had the same idea :) But is it fully supported by OC X?

dartcafe:
Funny, when you sit there and you are manually working on this and recognize afterwards, that the file you edited can be created fron occ.

dartcafe:
But it does not work as I expected. I have to call the migration manually via `occ migrate:execute`. I expected the creation of the database on enabling or installing the app. 

And OC goes it's own way. The migration is expeted in the appinfo folder, not in the lib folder. Manually copying it there causes an error because of unknown classes. I get mad. Maybe someone wants to make a fork and supports OC in it. My personal target is NC.

dartcafe:
I missed an entry in the migration table. Installation worked with NC13!

Steps to reproduce:
+ disable polls app
+ remove polls app
+ delete all tables starting with `oc_polls`
+ remove entry regarding in the table `oc_migrations`
+ extract this PR to the polls folder
+ enable app
voilà

Merging to develp-0.9?

splitt3r:
Let´s remove the empty methods and switch to the [Doctrine Types](http://www.doctrine-project.org/api/dbal/2.5/class-Doctrine.DBAL.Types.Type.html) the rest looks great.

dartcafe:
Done.

splitt3r:
Can you remove the use-migrations tag 😁 should only be needed for oC.

splitt3r:
Travis CI should work again now :)

dartcafe:
OK, I tested with the latest commits. Everything OK on my instance.

Issue #131, opened 2017-07-18
Can't save a TXT poll: "Click on cells to turn them green"


pmd2016:
If I try to add a public text poll, I get the following message:

> Nothing selected!
> Click on cells to turn them green

<img width="990" alt="screen_shot_2017-07-18_at_14_59_54" src="https://user-images.githubusercontent.com/19425130/28318279-760d957c-6bca-11e7-9cff-12433aef4802.png">

joergmschulz:
The alert exactly tells you what to do: In order to activate the selections you have to click on the red cells containing the X. Then it will work.

pmd2016:
Ah. I see. It's not that obvious what you need to do IMHO... 😉

v1r0x:
Can we close this or should there be a description or anything in the UI so it's clear what to do?

q00d0:
I had the same "problem". A description would be nice!

v1r0x:
Ok, as you wish 😉 Any suggestions?

pmd2016:
The way I interpret the 'insert text' screen is as answer options to the poll. It is not very clear that you have to answer the poll yourself before you can save the poll. It would be great if you can just setup a poll and let others answer without the need to answer the question yourself first. Does that make sense?

v1r0x:
You don't have to answer the questions, it's just about selecting the available questions you want in your poll. Only entries with a green checkmark will be in the poll.

q00d0:
Maybe radio buttons fix this - "active" and "inactive" or something like this.

DJCrashdummy:
IMHO this problem only occurs, if you haven't already created a date-poll with this app, so...

- a little bit more clear statement like in https://github.com/v1r0x/polls/issues/131#issuecomment-316059909 would definitely be fine.

- also activating the lines by default could be a solution.
if you've done a mistake, you start thinking about "how can i get rid of it" anyway. - and if you've already created a date-poll, the solution will be pretty obvious (with the same UI).

- if you really want to change the UI, maybe something like a trashcan for deactivating/deleting would make it more clear.

- or better, similar to https://github.com/v1r0x/polls/issues/131#issuecomment-316302033, the important thing is to have both options available to be more self-explaining _(if people have/see only one option, they usually don't think about it or any function behind it.)_:
wether it is the "X" and the "check mark" side by side and the selected gets colored, both with radio buttons or radio buttons with text is secondary.

pirate:
[...]

PR #260, opened 2017-12-03
Version 0.8.1

dartcafe:
Preparation for Version 0.8.1


splitt3r:
We should merge #252 first.

dartcafe:
Played a little bit with the line break in the description
This description 
grafik
gave this in the list
grafik
grafik
and this in the polls page
grafik

After the commit: 
grafik
grafik
grafik
grafik

dartcafe:
CRs applied

templates/goto.tmpl.php
```diff
@@ -81,7 +81,7 @@
 		$poll->getDescription() !== null &&
 		$poll->getDescription() !== ''
 	) {
-		$description = $poll->getDescription();
+		$description = str_replace(array("\r\n", "\r", "\n") , '<br/>', htmlspecialchars($poll->getDescription()));
```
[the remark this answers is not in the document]
dartcafe: Done

splitt3r:
Sorry. I think i will habe some time tomorrow to do some testing.

splitt3r:
The avatars are only applied after refreshing the page:

avatar

This looks a bit strange (deleted public poll in chrome / edge):

grey

Button moves to the right:

button

After copying the link a small notification would be nice sth. likle:

`OC.Notification.show(t('polls', 'Link copied to your Clipboard!'), {timeout: 5});`

We should fix this:

tooltip

All in all it looks great now :) we could also fix the small things later on.

dartcafe:
> The avatars are only applied after refreshing the page:

Strange. Here they are applied with no problem after posting. 
added_avatar

> This looks a bit strange (deleted public poll in chrome / edge):

Yup. We changed filename from polls.svg to app.svg. CSS updated

> Button moves to the right:

Create/Edit is the next project, so I will inspect later as well as #131

> After copying the link a small notification would be nice sth. like:

Changed this already in my test environment, but it was overwritten by poll. Added this again (stolen from files app 😆).

I also added a position option to the top positiond tooltips so it is diplayed under the items (I expected automatically positioning and discovered the option to define the position right now).

js/start.js
```diff
@@ -1,6 +1,50 @@
...
+			}
+		}, 3000);
+	});
+	clipboard.on('error', function (e) {
+		var $input = $(e.trigger);
+		var actionMsg = '';
+		if (/iPhone|iPad/i.test(navigator.userAgent)) {
+			actionMsg = t('core', 'Not supported!');
+		} else if (/Mac/i.test(navigator.userAgent)) {
+			actionMsg = t('core', 'Press ?-C to copy.');
```
splitt3r: This should also be ⌘-C shouldn´t it? We should only have this cliboard code once and share it between vote.js and start.js.

js/start.js
```diff
@@ -1,6 +1,50 @@
...
+			.tooltip('fixTitle')
+			.tooltip({placement: 'bottom', trigger: 'manual'})
+			.tooltip('show');
+		_.delay(function () {
+			$input.tooltip('hide');
+			if (OC.Share.Social.Collection.size() == 0) {
+				$input.attr('data-original-title', t('core', 'Copy'))
+					.tooltip('fixTitle');
+			} else {
+				$input.tooltip("destroy");
```
splitt3r: Could you replace the double quotes here 😁

js/start.js
dartcafe: I don't know, and didn't think about this too much. I just copied and pasted from files app. Actually we should think about creating more helper classes, to avoid code doubling. But I would prefer to do this in the next step.

js/start.js
dartcafe: Done.

js/start.js
splitt3r: Ok, i´m fine with that. But the question mark should be fixed before release. In vote.js it´s ⌘-C so it should be the same here.

dartcafe: Oops. Done

dartcafe:
The time is updated live. Just wait a few seconds.

splitt3r:
But isn´t it possible to trigger it after posting the comment?

splitt3r:
Without the `width: 100%;` it looks strange in Nc 13 for me:

width

splitt3r:
@v1r0x should be fixed now :)

dartcafe:
Seems something changed between beta 1 aud beta 3. I will optimize the css after the 0.8.1 release to make this more lean.

dartcafe:
Any reviews ongoing?

v1r0x:
Yes, sorry for the long idle time. Thanks to @splitt3r  the time appears directly after sending the comment, but is in english ('now') although my language is set to german ('Gerade eben'). After a reload it is correctly translated.

Is it on purpose that the toggle icon for actions and comments is on the left side?

joergmschulz:
On creating a date poll, the date picker chose the current time as sample for the pickable times: 20:49 instead of 20:00
grafik

dartcafe:
@joergmschulz Yes the time picker makes nonsense. This is subject of redesign of the create page.

dartcafe:
@splitt3r I was wondering, that nobody mentioned the position of the details button before. 😁 It is just because the dynamycal positioning did not work as expected, when opening the sidebar. So i left it there for the moment. I will think it over later, if no one has an easy solution right now. 

The translation will work after merging to the master, I guess, when the transifex bot will run over it.

splitt3r:
:joy: i mentioned it somwhere but thought it was somehow intended.

Let's change now to just now. Sounds better :+1:

v1r0x:
> The translation will work after merging to the master, I guess, when the transifex bot will run over it.

Then I'm fine with merging. 🎉

joergmschulz:
With this date picker nonsense, we shouldn't merge. Which aspects should be retested manually?

dartcafe:
The problem with the datepicker is already present in the master branch. So the merge will have no effect regarding the picker. And this PR is big enough. We can test this with a new version of the datetimepicker after the merge, befor releasing. 

Quick view by me with version 2.5.14: this problem does not occur anymore.
xdan/datetimepicker#618

dartcafe:
@joergmschulz see #270

PR #266, opened 2017-12-11
fix-oc-avatars

dartcafe:
Fix for non accessible avatars in public polls, when using owncloud.

### DIFF
--- a/js/vote.js
+++ b/js/vote.js
@@ -55,9 +55,20 @@ function switchSidebar() {
 }
 
 $(document).ready(function () {
-	// count how many times in each date
 	new Clipboard('.copy-link');
+	// count how many times in each date
 	updateBest();
+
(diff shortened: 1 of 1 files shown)


js/vote.js
```diff
@@ -55,9 +55,20 @@ function switchSidebar() {
...
 $(document).ready(function () {
-	// count how many times in each date
 	new Clipboard('.copy-link');
+	// count how many times in each date
 	updateBest();
+
+	// Temporary hack - Check if we have Nextcloud or ownCloud with an anomymous user
+	var $hideAvatars = false;
+	if (!document.getElementById('nextcloud')) {
+		$product = 'ownCloud';
```
splitt3r: $ indicates that the variable holds a jQuery object and this variable seems not to be used anywhere. And we should use single quotes in JS and PHP where possible (next line). Besides from that nice hack :D.

PR #269, opened 2017-12-15
Add -webkit- prefix for mask

zauguin:
Chrome requires a `-webkit-` prefix for the CSS `mask`.


dartcafe:
Updated in develop branch.

PR #270, opened 2017-12-19
datetimepicker.js Update to version 2.5.14

dartcafe:


splitt3r:
Let´s move the dependencyto an package.json and pass the oC locale to it like https://github.com/xdan/datetimepicker#datetimepicker.

joergmschulz:
something seems to be missing:
grafik

dartcafe:
The vendor dir was excluded in .gitignore, so the css did not make it's way into the repo. Fixed...

dartcafe:
@splitt3r This sounds like bohamian villages to me. 😀 How is this done?

dartcafe:
On the other hand, I would like to replace this datepicker with the one in core. The time picker can be omittet after redesign.

splitt3r:
But the one in core has no time selection?

splitt3r:
If we would introduce an ui like the doodle one, the core datepicker / timepicker would work perfectly fine 💪.

dartcafe:
No, we will make it better. We will not copy doodle. I want to see them copying us.

PR #273, opened 2017-12-20
Spelling: "owner", Log in, does not expire, your name

comradekingu:


templates/goto.tmpl.php
```diff
@@ -334,7 +334,7 @@
 			<?php if (User::isLoggedIn()) : ?>
 				<div class="notification">
 					<input type="checkbox" id="check_notif" class="checkbox" <?php if ($notification !== null) print_unescaped(' checked'); ?> />
-					<label for="check_notif"><?php p($l->t('Receive notification email on activity')); ?></label>
```
dartcafe: We should keep this short for preventing line break in the mobile view.

splitt3r: What about `Receive notifications per email`?

comradekingu: Added another commit, and simplified this. :)

splitt3r:
Looks good to me now 👍 what do you think @dartcafe?

PR #292, opened 2018-01-25
Change develop-0.9 to master

dartcafe:


dartcafe:
@splitt3r, @v1r0x Wouldn't it be easier to merge this into the master branch and continue developing from there?

splitt3r:
Hm the master should be stable but this isn´t 😕

dartcafe:
OK. What is missing. It is stable on my test instance... ? My thought is, that it could be more easy to merge now, instead of merging later. 

develop-0.9 should be even with master besides your optimations. master is even with the last release.

dartcafe:
This are the different files
grafik

splitt3r:
Mergeing should be no problem as every change will go into develop-0.9 and if version 0.9 is ready it will be merged into master. The branch stable-0.8 won´t get any new featuers that need to be merged into master only backports / fixes. At least that´s how it was intended 😁

appinfo/info.xml
```diff
@@ -16,12 +16,12 @@
     <category>tools</category>
     <website>https://github.com/nextcloud/polls</website>
     <bugs>https://github.com/nextcloud/polls/issues</bugs>
+    <use-migrations>true</use-migrations>
```
splitt3r: @dartcafe do we need this? I think the migrations work without it. Only oC uses that. Travis fails because this tag is not listed in https://apps.nextcloud.com/schema/apps/info.xsd.

dartcafe: Ah. Ok. Thanks for the hint.

dartcafe:
Right. But transifex makes it a little bit difficult. In the past there have been some conflict merging big PRs, because of changes in the l10n files. We will see.

Issue #294, opened 2018-01-26
Cannot delete my own poll


xylo:
### Steps to reproduce
1. Create a poll
2. Open the poll
3. Go to poll settings
4. Click on "delete poll" (german: "Umfrage löschen")
5. Click on "OK" when the confirmation dialog appears

### Expected behaviour
I expect the poll to get deleted.  Moreover there should be some sort of server communication.

### Actual behaviour
Except the confirmation dialog disappears nothing else happens when I click on "OK".  There is not even a communication with the server.

### Server configuration

**Operating system:** Linux 2.6.32-042stab127.2 #1 SMP Thu Jan 4 16:41:44 MSK 2018 x86_64

**Webserver:** Apache/2.4.25 (Debian) SVN/1.9.5 mod_fastcgi/mod_fastcgi-SNAP-0910052141 mod_jk/1.2.42 PHP/5.6.30-0+deb8u1 mod_python/3.3.1 Python/2.7.13 OpenSSL/1.0.2l (cgi-fcgi)

**Database:** mysql 10.1.26

**PHP version:** 5.6.30-0+deb8u1
Modules loaded: Core, date, ereg, libxml, openssl, pcre, zlib, bcmath, bz2, calendar, ctype, dba, dom, hash, fileinfo, filter, ftp, gettext, SPL, iconv, mbstring, pcntl, session, posix, Reflection, standard, shmop, SimpleXML, soap, sockets, Phar, exif, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xml, xmlreader, xmlwriter, zip, cgi-fcgi, PDO, curl, gd, gmp, imap, json, ldap, mcrypt, mysql, mysqli, OAuth, pdo_mysql, pdo_pgsql, pdo_sqlite, pgsql, readline, sqlite3, xmlrpc, xsl, mhash, Zend OPcache

**Nextcloud version:** 12.0.5 - 12.0.5.3

**Updated from an older Nextcloud/ownCloud or fresh install:** 

**Where did you install Nextcloud from:** nextcloud.com

**Signing status:** 

No errors have been found.

**List of activated apps:**

Enabled:
  - activity: 2.5.2
  - calendar: 1.5.7
  - contacts: 2.0.1
  - dav: 1.3.1
  - federatedfilesharing: 1.2.0
  - federation: 1.2.0
  - files: 1.7.2
  - files_sharing: 1.4.0
  - files_texteditor: 2.4.1
  - files_trashbin: 1.2.0
[...]

PR #295, opened 2018-01-26
Fix delete poll

dartcafe:
fix for #294


dartcafe:
Changed target to master, so we can downstream the fix after merge to the child branches.

splitt3r:
This should go into 0.9 and be backported to 0.8 stable. So we can release a version 0.8.2 with the fix and once 0.9 is stable it will also be in 0.9.

splitt3r:
We could also delete the master branch and do it like e.g. the server repo.

PR #296, opened 2018-01-26
Fix delete poll

dartcafe:
#294


dartcafe:
I will close this one and merge from master after pr #295 into master.

PR #315, opened 2018-02-21
[WIP] introduce vue

dartcafe:
Start of introducing vue.js for create/edit page


v1r0x:
Thanks for starting the vue port! :tada: 
After playing around with vue for another project, I think/hope I can contribute some ideas/code. Is it ok if I push to this PR? As a first step I'd move the components to signle file components. Thus, we can get rid of the string templates. ;)

dartcafe:
Yes, of course. Feel free to contribute to the app you started. 🤣 I was away last week (learned a lot in the silicon village in my high age), so my time is a little bit limited in the moment. 

Do you plan to contribute to the create template or the others? In the first case, we have to coordinate, because I am still trying in my test instance, and we should prevent to drift away.

dartcafe:
What would be the best way to load the components from their own files? `include_once`, `require_once` or...? I want to avoid using webpack and import everything into one file at this point.

splitt3r:
You can have a look at https://github.com/nextcloud/notifications/tree/master/js-src. I think we need Webpack anyway to bundle and compress everything.

dartcafe:
Urks... More frameworks. 😩

dartcafe:
A brief snapshot of todays status.
grafik
grafik

splitt3r:
Looks pretty nice already :+1: . I think at least the type selection should be a dropdown field.

dartcafe:
Hmmm. Would be one more click to switch.

splitt3r:
But yeah but i think checkboxes are only for sth. like yes / no with two options to choose one.

dartcafe:
@splitt3r Comming back to the drop down field. The type selection has only two options.  :confused:

splitt3r:
Lets try it your way. We can change it later on anyway.

dartcafe:
This changes are commited to #337 
The behaviour of the sidebar was annoying tome a long time. I made some changes to the layout and gave the controls and buttons space inside the breadcrump, so they are always accessible. 

In the end I am more happy with the result, even in mobile view.

bildschirmfoto vom 2018-05-23 um 08 25 07
bildschirmfoto vom 2018-05-23 um 08 25 12
bildschirmfoto vom 2018-05-23 um 08 25 20
bildschirmfoto vom 2018-05-23 um 08 25 15

PR #327, opened 2018-04-11
Create a android or ios poll application as doodle app

Ludovicis:
Hello, 

Can you create a android or ios poll application as doodle app, polleverywhere, or thepollapp , please ? (From your poll nextcloud app obviously :))

I think using a phone application is better than using the web browser on phone...

Thanks for your answer :)


splitt3r:
Please open an issue and not a pull request.

PR #337, opened 2018-05-10
Vue create webpack

dartcafe:
I switched to webpack. Please review, because, I just started with webpack.


splitt3r:
Souns good 👍 will have a look at it.

package.json
```diff
@@ -0,0 +1,44 @@
...
+    "babel-core": "^6.26.0",
+    "babel-loader": "^7.1.2",
+    "babel-preset-env": "^1.6.0",
+    "babel-preset-stage-3": "^6.24.1",
+    "cross-env": "^5.1.3",
+    "css-loader": "^0.28.8",
+    "file-loader": "^1.1.6",
+    "vue-loader": "^13.7.0",
+    "vue-template-compiler": "^2.5.16",
+    "webpack": "^3.6.0"
```
juliushaertl: Any reason you not use the latest 4.x version of we pack? :wink:

dartcafe: Yes. I don‘t like it. 😄 
Seriously, it is because, I am totally new to it. So any review is strong welcome. I will look after it and update.

dartcafe: Hmmm... 
No I get " webpack.optimize.UglifyJsPlugin has been removed, please use config.optimization.minimize instead."

Note to myself: Don't change a running system. 😈

juliushaertl: Maybe https://github.com/nextcloud/deck/pull/459/files will help you a bit, as I did the same version upgrade last week. ;)

dartcafe: Hmmm... To much to understand for me right now between coffee and bread. I reverted the commit and will make the update in another PR. Or maybe someone else can repair the configuration.

BTW: I get the Error "Error: webpack.optimize.UglifyJsPlugin has been removed, please use config.optimization.minimize instead".

dartcafe:
I reverted to webpack 3, because of some symptoms with webpack 4, I cannot handle right now. I suggest changing to 4 in another PR. 

For now, I would like to merge this PR, if there are no complains. #315 is rather big and old, so we should finish this work in #315 and continue after fixing all bugs in new PRs.

Issue #424, opened 2018-11-15
Database schema is invalid?


nickvergessen:
### What is going wrong?
**Describe the bug**
Database error when voting on a text poll

**To Reproduce**
1. Create a text poll
2. Add two options
3. Try to vote

**Expected behavior**
Vote should be saved,
Instead an exception is thrown by the database

### Information about your polls installation
**Polls version?** git/master

**Fresh installation or update from a prior version (from which one)?** new

**How did you install this version?(Appstore or describe installation)** git

### Information about your Instance of Nextcloud/ownCloud
**Nextcloud or ownCloud?**

**Which Version?** 14.0.3

### Server configuration

**Database:** MySQL 5.7.24

**PHP version:** php 7.1.24

#### Nextcloud log (data/nextcloud.log)
```
Nextcloud
Internal Server Error
The server was unable to complete your request.

If this happens again, please send the technical details below to the server administrator.

More details can be found in the server log.

Technical details
Remote Address: 127.0.0.1
Request ID: TAXTQr16P1mc3jjMsIj3
Type: Doctrine\DBAL\Exception\DriverException
Code: 0
Message: An exception occurred while executing 'INSERT INTO `oc_polls_votes`(`poll_id`,`user_id`,`vote_option_text`,`vote_answer`) VALUES(?,?,?,?)' with params [1, "test1", "test", "maybe"]: SQLSTATE[HY000]: General error: 1364 Field 'vote_option_id' doesn't have a default value
/home/nickv/Nextcloud/14/server/3rdparty/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
Line: 115

Trace
#0 /home/nickv/Nextcloud/14/server/3rdparty/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php(128): Doctrine\DBAL\Driver\AbstractMySQLDriver->convertException('An exception oc...', Object(Doctrine\DBAL\Driver\PDOException))
#1 /home/nickv/Nextcloud/14/server/3rdparty/doctrine/dbal/lib/Doctrine/DBAL/Statement.php(177): Doctrine\DBAL\DBALException::driverExceptionDuringQuery(Object(Doctrine\DBAL\Driver\PDOMySql\Driver), Object(Doctrine\DBAL\Driver\PDOException), 'INSERT INTO `oc...', Array)
[...]

PR #425, opened 2018-11-16
Consolidate and fix migration

dartcafe:
#424


lib/Migration/Version009000Date20181116063745.php
```diff
@@ -0,0 +1,69 @@
...
+use OCP\IConfig;
+use OCP\Migration\SimpleMigrationStep;
+use OCP\Migration\IOutput;
+
+/**
+ * Adding timestamp to options table
+ */
+class Version009000Date20181116063745 extends SimpleMigrationStep {
+	/** @var IDBConnection */
+	protected $connection;
```
nickvergessen: not used, you can delete it

lib/Migration/Version009000Date20181116063745.php
```diff
@@ -0,0 +1,69 @@
...
+
+	/** @var IConfig */
+	protected $config;
```
nickvergessen: not used, you can delete it

lib/Migration/Version009000Date20181116063745.php
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * @param IDBConnection $connection
+	 * @param IConfig $config
+	 */
+	public function __construct(IDBConnection $connection, IConfig $config) {
```
nickvergessen: method not needed (after you deleted the unused members), you can delete it

lib/Migration/Version009000Date20181116063745.php
```diff
@@ -0,0 +1,69 @@
...
+		$this->connection = $connection;
+		$this->config = $config;
+	}
+	/**
+	 * @param IOutput $output
+	 * @param \Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @since 13.0.0
+	 */
+	public function preSchemaChange(IOutput $output, \Closure $schemaClosure, array $options) {
```
nickvergessen: method not needed (because the parent class does exactly this), you can delete it

lib/Migration/Version009000Date20181116063745.php
```diff
@@ -0,0 +1,69 @@
...
+		return $schema;
+	}
+
+	/**
+	 * @param IOutput $output
+	 * @param \Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @since 13.0.0
+	 */
+	public function postSchemaChange(IOutput $output, \Closure $schemaClosure, array $options) {
```
nickvergessen: method not needed (because the parent class does exactly this), you can delete it

dartcafe:
> So in theory your files should be called `0009`

OMG. You're right.

dartcafe:
I can change this, but does this affect the migration system, if users installed the RCs?

nickvergessen:
Since RC is git only, you can do it.
Also your migrations look save enough to not cause troubles, because you check if the table exists before adding it again etc.

But you can simply change the file+class names and run the update locally and see if it does :boom:

dartcafe:
@nickvergessen thanks for help
- updated function names
- remove unused code
- consolidated migrations

lib/Migration/Version000009Date20180202213017.php
```diff
@@ -42,16 +42,16 @@ class Version009000Date20180202213017 extends SimpleMigrationStep {
 	public function changeSchema(IOutput $output, \Closure $schemaClosure, array $options) {
 		/** @var ISchemaWrapper $schema */
 		$schema = $schemaClosure();
-		if ($schema->hasTable('polls_dts')) {
+		if ($schema->hasTable('polls_dts') & $schema->hasTable('polls_options')) {
```
nickvergessen: You mean &&
And I would not check the second table here

lib/Migration/Version000009Date20180202213017.php
```diff
@@ -42,16 +42,16 @@ class Version009000Date20180202213017 extends SimpleMigrationStep {
...
 			$schema->dropTable('polls_dts');
 		}
-		if ($schema->hasTable('polls_txts')) {
+		if ($schema->hasTable('polls_txts') & $schema->hasTable('polls_options')) {
```
nickvergessen: Same

lib/Migration/Version000009Date20180202213017.php
```diff
@@ -42,16 +42,16 @@ class Version009000Date20180202213017 extends SimpleMigrationStep {
...
 			$schema->dropTable('polls_txts');
 		}
-		if ($schema->hasTable('polls_particip')) {
+		if ($schema->hasTable('polls_particip') & $schema->hasTable('polls_votes')) {
```
nickvergessen: Same

lib/Migration/Version000009Date20180202213017.php
```diff
@@ -42,16 +42,16 @@ class Version009000Date20180202213017 extends SimpleMigrationStep {
...
 			$schema->dropTable('polls_particip');
 		}
-		if ($schema->hasTable('polls_particip_text')) {
+		if ($schema->hasTable('polls_particip_text') & $schema->hasTable('polls_votes')) {
```
nickvergessen: Same

lib/Migration/Version000009Date20180202213017.php
dartcafe: Yes. I meant &&. Thanks. I thought it would be a good idea to check the existing of the new tables before deleting the old one. Just to prevent data loss. 

But maybe you are right. Migration seems to work properly here.

dartcafe:
> And you file names should be 0009 not 000009. But not too important.

@nickvergessen I am a nitpicker. :-) So thanks for the hint. I updated the file names.

dartcafe:
> Since you never released this yet, consolidating is okay. But in theory you should not do this as soon as it was released. You should also never edit a released migration, just add a new one.

@nickvergessen This was indeed because this was never released. I just wanted to make it more handy for the initial use of the migration system.

nickvergessen:
👍🏼

Issue #420, opened 2018-11-15
Start with "?" state instead of "x"


jancborchardt:
Currently when you go to a poll, it looks like this:
![polls now](https://user-images.githubusercontent.com/925062/48540093-6c65ba00-e8b9-11e8-922a-c8efd83622cb.png)

That’s strange because it seems like the app already submited a decision for you. Instead we should start with the "Undecided / if need be state", since it also is a call to action with the "?" icon.

![polls start with unsure](https://user-images.githubusercontent.com/925062/48540092-6bcd2380-e8b9-11e8-9d47-1d2cfe3c07d2.png)

dartcafe:
> That’s strange because it seems like the app already submited a decision for you.

Difficult discussion, which we already had. Some say start with maybe, some say (including me), force clear answers and say no to every option, the voter does not state his clear optinion.

DJCrashdummy:
i'm also a fan of clear answers (and don't like the maybe-option at all), but as long as it is an possible option it **should definitely be used as initial state** to call for an action or **if the participant hasn't changed it's state** (whether intentional or by accident) **to correctly indicate that**.
also for polls with a lot of options, which have to be clarified one by one (e.g. date-polls) it is also useful for the participant to easily see which of the options he hasn't already processed.

DJCrashdummy:
i also thought about the possibility (while creation of a poll) to enable/disable a strict "yes/no"-policy...

thinking about the whole possibilities:
- the **yes (green)** and **no (red)** is out of question...
- but the **initial state** should be a neutral **question mark** with a neutral colour (perhaps a colour to catch the participants attention)
--> probably if a state wasn't touched also post it like that. or remember the participant that a/some decision(s) has to be made. - or force him to make all decisions, if the creator has chosen this.
- and then a maybe/**"if need be"** may also be reasonable: a **check mark in brackets** seems good to me (also with a neutral colour or perhaps a very light green or greenish colour, since it is rather yes than no)
...because a question mark is not really an answer, not even in a poll!

the biggest question for me in this case (except the strict yes/no-policy) is, if should it be possible to get back to the initial state once touched.

jancborchardt:
Coming up with our own logic would be much too confusing. We should simply adapt what Doodle is doing for now as they are clearly the most used solution in this field.

dartcafe:
So you vote vor:
- default is "no",
- display "No" without a red cross
- display "maybe" as a checkmark in brackets?

[...]

PR #435, opened 2018-12-01
Change actions in vote page

dartcafe:
#420


dartcafe:
@jancborchardt First step. Comments? 
bildschirmfoto vom 2018-12-30 um 17 31 05

jancborchardt:
Nice! Some feedback:

- I’d drop the shadow of the checkboxes, it’s not really conforming to Nextcloud style
- Also the boxes don’t all need to have blue border. Color needs to convey meaning and here it doesn’t. Normal boxes should have dark border, and the selected ones either green or yellow (depending on their state).
- If the list of participants is long, the vote button will not be findable. It should be near to what you are voting on. Ideally at the end of the list on the right.
- The "Receive notification email on activity" could go into the 3-dot menu where we also wanted to put "Toggle all". (Can’t see the "Toggle all" anymore though.)
- The "New" field looks a bit off. I guess this will only show when you are editing your vote and new options were added? Maybe just having an empty checkbox is simpler? How does Doodle handle this?
- Similar for the "This poll was updated since your last visit. Please check your votes". Is it necessary? For long lists of voters it will not be visible there on the bottom. If updates happened, maybe it makes more sense to only show the cells in edit mode where changes were made, like dates were added etc? That makes it more obvious you need to take action.

bpcurse:
Looks good! I would agree with @jancborchardt on his aspects, adding my two cents:

> The "Receive notification email on activity" could go into the 3-dot menu where we also wanted to put "Toggle all". (Can’t see the "Toggle all" anymore though.)

I see the point (and like it in theory) but as this is ok for a regular user, to a new user it is probably more intuitive when it is not hidden in a menu?

> Similar for the "This poll was updated since your last visit. Please check your votes". Is it necessary? For long lists of voters it will not be visible there on the bottom. If updates happened, maybe it makes more sense to only show the cells in edit mode where changes were made, like dates were added etc? That makes it more obvious you need to take action.

What about this way to display it?

polls_mockup

dartcafe:
A little mockup:
bildschirmfoto vom 2018-12-30 um 21 55 07

@jancborchardt 
> * I’d drop the shadow of the checkboxes, it’s not really conforming to Nextcloud style
> * Also the boxes don’t all need to have blue border. Color needs to convey meaning and here it doesn’t. Normal boxes should have dark border, and the selected ones either green or yellow (depending on their state).

The shadow was a try to get a click down effect. I changed the colors according to their state. Or any suggestion for another color?

> * The "Receive notification email on activity" could go into the 3-dot menu where we also wanted to put "Toggle all". (Can’t see the "Toggle all" anymore though.)

Hmm, long text for this, I will try.

> * The "New" field looks a bit off. I guess this will only show when you are editing your vote and new options were added? 

Yes right. The intention is, f the poll changed after a user voted, he is informed, which options were added. I think it is important.

> Maybe just having an empty checkbox is simpler? How does Doodle handle this?

Empty means no, as doodle does. See discussion in #420. I didn't check the behavior due to new options.

> Similar for the "This poll was updated since your last visit. Please check your votes". Is it necessary?

This is just an additionally information about the new options. 

@bpcurse Good idea about the position. But I will change this, when we migrated this to vue also. At the moment it depends on the order of the code, so I have this information after the table is build. 

The mixture of PHP and JS is not very nice. Having PHP for the backend and vue for frontend is much more convenient. For now, I wanted some prior discussion for the design, prior to the vue migration.

jancborchardt:
@dartcafe very very nice mockup, looks great! :)

> The shadow was a try to get a click down effect. I changed the colors according to their state. Or any suggestion for another color?

Cool, that seems fine. The click down effect is not needed. I’m only worried about the yellow color as it’s very light. Did you take the colors from the `core/css/variables.scss`? Ideally we take a darker yellow / orange for better visibility.

> Hmm, long text for this, I will try.

We can cut it to "Notification mails" or "Receive notifications".

> Yes right. The intention is, f the poll changed after a user voted, he is informed, which options were added. I think it is important.

Ok, makes sense. We can still see how this works in actual use. :)

-------------

As for the 

> **This poll was updated since your last visit. Please check your votes**

We should just put it in bold directly below the description text. That way it’s not inbetween the other interface elements.

And lastly, the "Vote" checkmark button should ideally be a circle, minimum 44*44px clickable area.

What do you think? :)

dartcafe:
> We should just put it in bold directly below the description text. That way it’s not inbetween the other interface elements.

Yes. See my comment to bpcurse. This is much more easier after the migration to vue. 

> Did you take the colors from the `core/css/variables.scss`?

I could use `--color-sucess`, `--color-error` and `--color-warning`. But for the background I am missing ideas. Can I blur it (calculated from the vars)?


> The click down effect is not needed.

I liked it, because, I manged to do this. :-) 

> And lastly, the "Vote" checkmark button should ideally be a circle, minimum 44*44px clickable area.

Yeah, like the other icon buttons too. Seems that just a div with classes `actions creatable` are missing.

jancborchardt:
> But for the background I am missing ideas. Can I blur it (calculated from the vars)?

How about the "lighten" color function? :) So `background-color: lighten(var(--color-success));` - not sure if the syntax is correct, am mobile atm. :)

dartcafe:
> How about the "lighten" color function?

As far, as I see, this is a SASS function and does not work with the css variables.

`background-color: lighten(var(--color-success), 50%);` does not work.

jancborchardt:
🔮 Summoning SCSS/CSS wizard @skjnldsv

💨💥💨

🧙 Master, do you know a solution here? 🙏

skjnldsv:
Indeed, you cannot use a scss function on css vars.
Why do we need to lighten here? Are the default colours too vibrant?

jancborchardt:
> Why do we need to lighten here? Are the default colours too vibrant?

Check the screenshot at https://github.com/nextcloud/polls/pull/435#issuecomment-450587141
- The color itself is used for the icon, that is the ✔ or the ❌
- For the background of the cell we should use a lightened version of the color, just like Doodle.

skjnldsv:
Well, the idea of using css-variables is to have a way to be compatible with the dark theme as the vars can be changed per user.
So the real question is: shall we change the colours if there is a dark theme? If not then use scss variables for those colours :)

dartcafe:
Moving the button has consequences on mobile view, so I leave it for later.

PR #440, opened 2018-12-11
Update max-version to current dev version 16

jancborchardt:
This is necessary to make the Polls app work on the current master / dev version of Nextcloud which is version 16. :)

Please review @dartcafe @splitt3r @v1r0x

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -23,6 +23,6 @@
     <screenshot>https://raw.githubusercontent.com/nextcloud/polls/master/screenshots/vote.png</screenshot>
     <screenshot>https://raw.githubusercontent.com/nextcloud/polls/master/screenshots/edit-poll.png</screenshot>
     <dependencies>
-        <nextcloud min-version="14" max-version="15" />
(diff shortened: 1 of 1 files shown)


jancborchardt:
@dartcafe @v1r0x we should also have a release, so it works on Nextcloud 15 via the app store? https://apps.nextcloud.com/apps/polls :)

v1r0x:
> Haha! :D You should sync it with some software maybe? wink Anyway, @rullzer @MorrisJobke how can @v1r0x get a new one?

Good idea, but can't think of a reliable and nice software to do this... :stuck_out_tongue_closed_eyes: 
Already have a PR for a new key: https://github.com/nextcloud/app-certificate-requests/pull/205 maybe everyone is busy wrapping christmas gifts :)

dartcafe:
> Good idea, but can't think of a reliable and nice software to do this...

What about Keybase?

jancborchardt:
I think there’s also a thing called "Boxdrop" or so I heard …

Issue #445, opened 2018-12-14
0.9.1: Vote options view is distorted (many dates, browser window too small to view them all)


herrmannsdorfer:
The vote options view is distorted with rel. 0.9.1 and NC 15.0.0 (did not test with NC 14).

As soon as there are more vote options than space in the browser window the options get compressed into that view. The heading remains intact and needs to be scrolled.

Tested with browsers on macOS and Win (Safari 12, FF 64, Chrome 71, IE 11).

Example with 31 options and only 10 fitting in the browser space:

![bildschirmfoto 2018-12-14 um 17 03 12](https://user-images.githubusercontent.com/43957716/50015012-c8f5eb00-ffc5-11e8-9f02-e9fb939550fd.png)

PR #446, opened 2018-12-15
fix poll width

dartcafe:
#445


dartcafe:
grafik

herrmannsdorfer:
Tx for the quick fix!

PR #476, opened 2018-12-31
Update environment

dartcafe:
Change to dev environment
* added npm build process to Makefile
* renamed create-poll.js to polls.js
* added eslint
* prepared jest
* updated dependencies
* fixed some typos in file names (lowercase -> camelcase)


dartcafe:
@v1r0x I made some changes to the environment, the file structure and the Makefile. Tests were positive on my build environment. Can you check, if everything works on your side?

Makefile
```diff
@@ -6,85 +6,107 @@
...
+
+lint-fix:
+	npm run lint:fix
+
+# Removes the appstore build and compiled js files
 .PHONY: clean
 clean:
-	rm -rf ./build
+	rm -rf $(build_dir)
+	rm -f ./vendor
```
v1r0x: Either use `-rf` since vendor is a folder or remove that line. I don't see a point, why the php deps should be deleted?

dartcafe: Good question. I thought it is a good idea to make a clean dev environment and pull the fresh libraries. Anyway the composer seems not make sense, because the tests fail on my instance due to a missing base.php. Your decision. :-)

v1r0x: I'd drop the `clean` target from the `appstore` target, because all files/folders in `clean` are regenerated in `appstore`. If anyone wants a clean repo, they can run `make clean && make appstore`.

dartcafe: Hmmm. Maybe we should change the building commands another way: 

see last commit

In the end 
- `clean-dev:` prepares a fresh `dev-setup`
- `clean:` prepares for a new `appstore` build without the need for a complete new `dev-setup`
- `appstore:` just runs a new appstore build with compiling js
- `all:` runs `dev-setup` and `appstore` 

What do you mean

dartcafe: @v1r0x Any more comments on this, or may I merge?

v1r0x: Sorry, everything's fine :)

PR #483, opened 2019-01-07
Update owncloud version

ho4ho:
> App developers should re-release their apps after setting the “max-version” field to “10” instead of “10.0” in “appinfo/info.xml”, and increase the app’s version as well
to make sure ownCloud picks up the change. This is required because else ownCloud 10.1 will think that the apps are not compatible any more and will refuse to update or enable them.

https://central.owncloud.org/t/owncloud-core-switching-to-semver-apps-need-re-release/17054

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -5,7 +5,7 @@
     <name>Polls</name>
     <summary>A polls app, similar to doodle/dudle with the possibility to restrict access.</summary>
     <description>A polls app, similar to doodle/dudle with the possibility to restrict access (members, certain groups/users, hidden and public).</description>
-    <version>0.8.3</version>
+    <version>0.8.3.1</version>
     <licence>agpl</licence>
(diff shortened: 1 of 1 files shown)


v1r0x:
Thanks for your contribution!

Unfortunately, oC is no longer supported. I updated the readme to clarify this. Sorry for the inconvenience!

ho4ho:
@v1r0x Thanks, but why not update existing https://marketplace.owncloud.com/apps/polls with small change here?

v1r0x:
I lost my signing keys and decided to not create/request a new one for oC.

ho4ho:
Sad :cry:

dartcafe:
@v1r0x @ho4ho I updated the info.xml directly in the branch, because there is no reason to not keep the stable-0.8 branch for further development. Maybe compatibilty comes back or someone wants to maintain this branch.

Issue #313, opened 2018-02-16
Polls overview - Don't show red icon "you did not comment"


bpcurse:
At the moment a red comment icon is shown in the polls home / overview if the user has not commented the poll. Usually a red sign indicates something is wrong or input / action is needed, which seems misleading on behalf of comments.
Proposal: Only show the green (could then also be neutral black/white or grey) comment icon if a comment was made. Do not show the red icon at all.

dartcafe:
You are right. We should indicate the user specifig information in another way. Maybe just show the relevant indicators and hide them, if they do not apply.

joergmschulz:
we should display the red flag if the user hasn't polled yet. But for the comment, a red flag shouldn't be displayed, in my opinion. The green one might help users.

Issue #228, opened 2017-10-30
Move list of polls into left .app-navigation, use .app-content exclusively for poll content


jancborchardt:
As mentioned in https://github.com/nextcloud/polls/pull/166#issuecomment-335975706

> This list of polls is what the left app-navigation is made for. :) We don’t really need a complex table to list these polls, it’s unnecessarily confusing. The navigation can have a simple section for Open polls first and Closed polls below, show the icon for if you voted or not, show the sharing indicator for if it’s shared with users or shared via link, and other details can go in the 3-dot menu. :)

> The left app-navigation is for listing the polls, and the app-content is for displaying one poll. :)

> Currently the Polls app has a mode in the content: Either list of polls, or one poll. That is confusing. The Tasks app only ever shows tasks in the content.

> Tasks app uses app-navigation for a list of task lists, Polls would use it for a list of polls. Tasks uses app-navigation for a list of the tasks in the selected list, Polls would use it for a list of the participants and answers in that selected poll. (The poll detail view as we have now already.)

>  For the empty main view at the start we will have the proper .emptycontent view, looking something like

> Polls app icon
> No polls created yet
> [ Start with your first poll ]

> Just as in all other apps, and in empty folders etc. And as @skjnldsv mentioned, when you open the app, the first poll should be loaded. Or rather, the last one you had open.

> What do you think @v1r0x @skjnldsv @splitt3r @dartcafe? The main point is that the two entirely different view modes in the content are very confusing. This is the same issue we have and will fix in the Deck app: nextcloud/deck#12

dartcafe:
I thought over this a lot of times and I am not happy with the navigation bar for the polls list. When there are many polls, the navigation can mutate to a huge list of polls. And there will be missing details for organisation, which can be important. Finding the right poll on a list with details can be more easy, than finding it via the tight list on the left. 

Additionally the app navigation can get overloaded, if there are many polls in the list and then the details are important to find the right poll together with quick filtering.

I believe, that the first poll is not the one, which interests me. Normally a user enters the app to manage the polls or to vote. When he wants to vote, he wants to enter the poll, he is currently affected by. 

I would prefer the app navigation being for filtering and settings. 

possibly interesting filters can be:
- Polls created by me
- open polls
- Polls, in which I voted
- Polls which are expired
- Poll which expire soon
- Archived polls
- ...

dartcafe:
done in https://github.com/nextcloud/polls/tree/dev-1.0

[...]

PR #487, opened 2019-01-10
Polls List to Vue

dartcafe:
Migrating the list of polls to vue
 #313 Done as well


dartcafe:
@v1r0x I changed the icons to the icon type. What do you think?

bildschirmfoto vom 2019-01-12 um 18 50 01

dartcafe:
Update. Ready for Test and review... 
bildschirmfoto vom 2019-01-13 um 14 04 33

lib/Controller/ApiController.php
```diff
@@ -465,12 +470,36 @@ public function getPolls() {
...
-			$eventsList[$eventElement->id] = $this->getEvent($eventElement->id);
+			$eventsList[] = $this->getPoll($eventElement->id);
 		}
 
 		return new DataResponse($eventsList, Http::STATUS_OK);
 	}
 
+	/**
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
```
juliushaertl: There should be a CSRF check on this route.

templates/create.tmpl.php
```diff
@@ -24,5 +24,5 @@
 
 ?>
 
-<div id="create-poll"></div>
+<div id="app-polls"></div>
```
juliushaertl: I guess you can get rid of that template file :wink:

dartcafe:
> 
> 
> Good work porting @dartcafe! :)
> 
> I still think that having a whole page for these details is not really necessary. It would work really nicely (and also fit better within Nextcloud) if the polls list would simply be the left #app-navigation: https://docs.nextcloud.com/server/15/developer_manual/design/navigation.html#app-navigation-menu
> 
Remember #228? 😏
I think we should differentiate based on the user's role.
- For adminitrative cases (i.e. site administrators) this page is more useful, when there are sorting and/or filtering functions. 
- Someone who creates or manages own polls, needs navigation for managing own polls.
- Someone who only participates in polls needs help in finding polls, where his vote is wanted. 

So the navigation should know categories like
- Your current polls
- Your expired polls
- Your closed polls (tbd)
- Polls you voted in
- Polls you are invited to
- Public polls
- All polls
- ... 
 
The navigation will come, but first, I want to migrate to a complete vue app. This are much more easy after that step.

>
> - What are polls sorted by? I assume last activity? If not, that should be the case. :)

Hey. Don't push me. This was the first step. :smile:

> * The icons use color at the moment which is not accessible. It’s also too much information. Normal (active) polls should have the normal (black) icons, and expired ones could be slightly greyed out for example. But it shouldn’t be only communicated via color.

I will try this
 
> * There is a green counter bubble on the icon which is non-standard, and should be used via the counter element https://docs.nextcloud.com/server/15/developer_manual/design/navigation.html#counter

Hmmm. The bubble should indicate, that there are comments. A simple counter could indicate anything. 

> * Menu goes into the menu ;) https://docs.nextcloud.com/server/15/developer_manual/design/navigation.html#menu – "Delete poll" should be last in the list as it is a destructive action
> * Creator or creation date is not really important in this overview. It’s enough to have this visible when you open the poll

Not for administation use.

> * Name of poll goes in first/primary row

Why? I would leave the type indicator in the first row. 
> * Access and expiration date can go in the secondary row – this is not documented yet but used by the Talk app.
> * "Participated" could be an overlay on the icon much like for the "Favorites" in the Talk app

Isn't it to much indication, if we keep the counter. Some design discussion is welcome 
> Also, the modal used seems to be non-standard?

Do we have one? @skjnldsv What about a modal vue component in nextcloud-vue?
> * The corners should be rounded
> * The header has a blue background, should not have any
> * The buttons are confusingly labeled, should be meaningful like "Keep poll" and "Delete poll"
> * Delete button should be right-aligned, keep button left-aligned.

This was a quick implementation before my holydays. I was happy I found a simple solution for a modal, whis is, to be honest, not really modal.

> Also cc @nextcloud/designers for feedback. :)

Yes, please!! 👋

dartcafe:
Ehm, OK. https://github.com/nextcloud/nextcloud-vue/pull/212 😆

lib/Controller/ApiController.php
dartcafe: @juliushaertl Umpf. Have the CSFR thing generally on my to do, because I have to understand how to deal with this. Any how to available?

templates/create.tmpl.php
dartcafe: > I guess you can get rid of that template file 😉

Yep.

dartcafe:
I get an error on compiling via npm. 
```
ERROR in ./src/js/router.js 30:9
Module parse failed: Unexpected token (30:9)
You may need an appropriate loader to handle this file type.
|
| var Create = function Create() {
>   return import('./views/Create');
| };
|
 @ ./src/js/main.js 25:0-30 65:10-16
```

Anybody can give me a hint? It worked before. @nextcloud/javascript

dartcafe:
@jancborchardt especially for you.... 😄 
bildschirmfoto vom 2019-01-29 um 17 50 48

The participation indicator is not satisfying. Any idea?

lib/Controller/ApiController.php
dartcafe: Didn't thik, that it was so easy.... 😆

dartcafe:
> 
> Too fast ! You don't even have time to think about it: BAM! Already here! ;)

I'll wait, until ist is in the release and will see, what it is able to do.

dartcafe:
Updated the participated indicator.
grafik

dartcafe:
@skjnldsv I will leave the counter for the moment as it is, because it is the default style of the server. If no more comments occur, I would like merge this and prepare for release.

PR #501, opened 2019-02-03
allow <br> in description

### LINKED ISSUE #500: Allow linebreaks in poll description and comments
It is not possible to add a newline in the output of the poll-description and in comments

If you add `<br>` it will shown as code there and if you add a line break, it are cut out completely of the output, not even replaced with a space, so the last character and the first of the new line glue together without space.

I can think of two solutions:

1. allow `<br>` in comments and the description 
2. or render new lines as a visible New line in the output
3. Allow markdown as [in announcementcenter](https://github.com/nextcloud/announcementcenter/pull/130/files)

I prefer the first or third solutions, the first is easy to achieve with PHP `strip_tags`, Example:
```
// allow <br> <p> and <a>:
echo strip_tags($text, '<br><p><a>');
```

rubo77:
closes #500


rubo77:
It is missing a solution for the comments, which also strip Newlines and show `<br>` as code.

can someone give me a hint, where to find the line that interprets the comments?

PR #502, opened 2019-02-03
Shift all dates of a date poll with given step and unit

dartcafe:
Allow shifting of all date options in one step.
shiftdates


v1r0x:
Like it! :sunglasses: 

I pushed a small fix, but beside that: Nice!

Issue #461, opened 2018-12-23
Polls with expire date could not be created/edited


nextgen-networks:
### What is going wrong?
**Describe the bug**
Polls created with an expire date are not created/could not be edited.

**To Reproduce**
use an poll with an expire date

**Expected behavior**
polls should be created/edited regardless if the utilize n expire date or not

### Information about your polls installation
**Polls version?** 0.9.5 / Fresh installation

**How did you install this version?(Appstore or describe installation)**
Appstore

### Information about your Instance of Nextcloud/ownCloud
**Nextcloud or ownCloud?**
NC

**Which Version?**
15.0.0

### Server configuration

**Operating system**: Ubuntu 16.04.5

**Web server:** Apache 2

**Database:** MySQL 5.5.27

**PHP version:** 7.2

**Signing status:** Ok

nextgen-networks:
thx for fast reply - I'll provide info in the next few days - now @ PoC buildup for 35c3

v1r0x:
@nextgen-networks you can provide your info at 35c3. I'll be there and (hopefully) working on polls 😉

nextgen-networks:
@v1r0x :-D
Great to see you in person at 35c3 - just give me a hint where you're located at when you've arrived or call dect 2069

v1r0x:
@nextgen-networks I'm either in heaven or running around doing angel stuff.

dartcafe:
I accont confirm. works here as designed on several instances. Some more information would help. 
Configuration, screen shots, ...

gmk1981:
Hi,

i have the same problem. When i check the expiration field, the poll could not be created. The status wheel rotates endlessly and nothing happens. 
Server: Ubuntu 16.04 
Nextcloud:Version: 14.0.4.2
PHP Version: 7.0.32
DB: mysql




![grafik](https://user-images.githubusercontent.com/42406145/50981235-d043e200-14fa-11e9-9299-7bbaeb88869d.png)

fnadeau:
I have the same issue. Works fine if no expiration is given.

A nice addition would be to have a warning on the web page that "something went wrong". The only thing I have is the working button. No info is given that it didn't work.

Would like to help more, but SQL and JS are not my strong suits. The mariadb setup says that the expire field is of datetime format and doc doesn't says it support the timezone in it: https://mariadb.com/kb/en/library/datetime/

so 2019-01-13T21:00:00.000Z doesn't look valid.

I hope it helps

[...]

PR #514, opened 2019-02-10
fix expire date could not be created/edited (#461)

dartcafe:
#461 - Polls with expire date could not be created/edited


andriusign:
Seems that expiration date/time does not equal timezone set in Nextcloud settings.

dartcafe:
Hmmm. It should. I have to check that.

dartcafe:
@andriusign Can you open another bug issue for that please?

Issue #511, opened 2019-02-06
No difference between hidden and open poll


HAVP:
All LDAP User can see hidden polls:

Same issue as described here:
https://github.com/nextcloud/polls/issues/101

dartcafe:
merged to master

PR #515, opened 2019-02-10
#511 - added missing access check

dartcafe:
#511 access permission was not properly checked


dartcafe:
Tested a few compinations, and will merge to get the master up to date.

Issue #532, opened 2019-02-25
cannot share poll (only share option)


cimbi:
### What is going wrong?

Since update to 0.10.0 I cannot share the polls with other users via Only shared option. There is simply no widget.
![poll_bug](https://user-images.githubusercontent.com/15217634/53339973-1c5eb080-3908-11e9-94b2-612301c8d7a2.png)

brtptrs:
I can confirm this problem

brtptrs:
![z_screenshot_20190225_003](https://user-images.githubusercontent.com/8102905/53351505-b0894180-3921-11e9-9b76-10a90da16cf4.png)

Firefox shows these errors when "apps/polls/new" is called.

gschbro:
Confirmed, same here :-(

horstvogel:
confirmed, same Problem on Nextcloud 15.04 with 0.10.0

dartcafe:
Ther seems to be a problem in the build process. I will check that. The compiled js in the package differs from my local compilation.

dieter-wilhelm:
I've the same problem on NextCloud 15.0.5

dartcafe:
https://github.com/nextcloud/polls/releases/tag/v0.10.1 should fix that. Appstore release will come up soon.
@v1r0x If you have time...

v1r0x:
@dartcafe Done! Thanks for your fixes!

horstvogel:
I have still the Problem, on Nextlcloud 15.05 with polls 0.10.1, even after a complet Restart of the system

gschbro:
Confirmed, same here NC 15.0.5 Polls 0.10.1 :-(

dieter-wilhelm:
The same here on Nextlcloud 15.05 with polls 0.10.1, I even had to clear my FF cache to see the old polls.  But still not able to share old and new polls.

dartcafe:
@v1r0x I had an issue with the build system on debian. Could you please make an `npm install` and `npm update` and run the build again. I can reproduce this issue but didn`t have it after updating the buld system.

Strange, I believed, npm would handle it by itself on `npm install`.

dartcafe:
Just for clarification:
* my windows build worked fine
* with my debian build (make all) I could reproduce the problem in the package.
* updating npm fixed the problem and the debian build was ok

andriusign:
Confirm, same error. Nextcloud 15.0.5, 
[...]

Issue #531, opened 2019-02-25
CSP err with 0.10.0


herrmannsdorfer:
I just updated polls from git and now I receive an CSP err and a blank screen opening polls:

![bildschirmfoto 2019-02-25 um 09 37 22](https://user-images.githubusercontent.com/43957716/53324513-3d5fdb00-38e1-11e9-9015-e70a1dc3ef93.png)

In Chrome this err reads as:

> Refused to execute script from 'https://files.herrmannsdorfer.de/nextcloud/apps/files/' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.

I updated from a working release 0.9.4 using
`git clone https://github.com/nextcloud/polls.git`

0.10.0 was not offered with web updater (maybe since I was cloning the repo the last couple of updates?).

brtptrs:
I get these CSP errors for 0.10.0 in Firefox. The upgrade was made normally via the update.

![z_screenshot_20190225_002](https://user-images.githubusercontent.com/8102905/53352240-b6cbed80-3922-11e9-851c-00964f97b478.png)

dartcafe:
This seems not to be polls related, but there have been some changes to the CSP options in the code. I will investigate.

dartcafe:
I cannot reproduce this. I tried this with different browsers. If you have more information...

herrmannsdorfer:
I deleted 0.10.0 from our apps directory. Then I downloaded and enabled 0.10.1 from app store. Polls is working again.

Dunno if 0.10.1 cured the issue or the clean install from the store. In the past I did several `git clone` in order to check RCs or PRs. Maybe that broke something.

dartcafe:
Good to read.

PR #542, opened 2019-03-04
Repckaging for #532

dartcafe:
To do:
- [x] Update release date in CHANGELOG.md before release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,15 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## [0.10.1] - tbd
+## [0.10.2] - tdb
+
+  - #532 - cannot share poll (only share option)
+
+## [0.10.1] - 2019-03-02
+
+### Fixed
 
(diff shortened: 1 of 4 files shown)


dartcafe:
@v1r0x Can you test the release before publishing, if #532 is fixed after update modules?

v1r0x:
Couldn't test it...CSP blocks fonts and doesn't load polls. Force-reload/private windows won't work either

dartcafe:
Ist it possible, that a browser extension causes this? i.e. vue-devtools?

v1r0x:
No, it's a new VM with only nextcloud installed. I can try another system tomorrow. Fingers crossed this works....Debugging is a nightmare in NC

dartcafe:
is it like #531?

v1r0x:
Yes, same problem as in https://github.com/nextcloud/polls/issues/531#issuecomment-467078847
I tried it on another machine and the build looks good. Disabled and removed the old polls version and enabled/installed the new build

dartcafe:
@v1r0x Would you add the release date to the readme and fire this up?

PR #596, opened 2019-06-04
Updated version and dependencies

dartcafe:
* security fix


dartcafe:
@v1r0x We need a new release due to an security update of axios (#593)

PR #672, opened 2019-12-10
Fix tests!

tortuetorche:
✔


lib/Controller/EventController.php
```diff
@@ -164,7 +164,7 @@ public function get($pollId) {
 			'fullAnonymous' => boolval($this->event->getFullAnonymous()),
 			'allowMaybe' => boolval($this->event->getAllowMaybe()),
 			'voteLimit' => $this->event->getVoteLimit(),
-			'showResults' =>$this->event->getShowResults(),
+			'showResults' => boolval($this->event->getShowResults()),
```
dartcafe: The type STRING was by intention and a preparation for another feature.

lib/Controller/NotificationController.php
```diff
@@ -166,7 +166,7 @@ private function sendNotifications($pollId, $from) {
 			}
 			$url = $this->urlGenerator->getAbsoluteURL(
 				$this->urlGenerator->linkToRoute('polls.page.vote',
-					array('hash' => $poll->getHash()))
+					array('token' => $poll->getToken()))
```
dartcafe: Makes no difference, since not used anymore. Will be deleted in next version.

lib/Db/Event.php
```diff
@@ -42,6 +42,8 @@
  * @method void setCreated(string $value)
  * @method string getAccess()
  * @method void setAccess(string $value)
+ * @method string getToken()
```
dartcafe: Makes no difference, since not used anymore. Will be deleted in next version.

lib/Db/Event.php
```diff
@@ -66,7 +68,7 @@ class Event extends Entity implements JsonSerializable {
 	protected $owner;
 	protected $created;
 	protected $access;
-	protected $hash;
+	protected $token;
```
dartcafe: Makes no difference, since not used anymore. Will be deleted in next version.

lib/Db/Event.php
```diff
@@ -89,8 +91,8 @@ public function jsonSerialize() {
 			'isAnonymous' => boolval($this->isAnonymous),
 			'fullAnonymous' => boolval($this->fullAnonymous),
 			'allowMaybe' => boolval($this->allowMaybe),
-			'voteLimit' => $this->showResults,
-			'showResults' => $this->showResults,
+			'voteLimit' => $this->voteLimit,
+			'showResults' => boolval($this->showResults),
```
dartcafe: The type STRING was by intention and a preparation for another feature.

lib/Migration/Version0010Date20190801063812.php
```diff
@@ -86,9 +86,9 @@ public function changeSchema(IOutput $output, \Closure $schemaClosure, array $op
 				]);
 			}
 			if (!$table->hasColumn('show_results')) {
-				$table->addColumn('show_results', Type::STRING, [
+				$table->addColumn('show_results', Type::BOOLEAN, [
```
dartcafe: The type STRING was by intention and a preparation for another feature.

lib/Migration/Version0010Date20190801063812.php
```diff
@@ -86,9 +86,9 @@ public function changeSchema(IOutput $output, \Closure $schemaClosure, array $op
...
 					'notnull' => true,
-					'lenght' => 64
+					'default' => 0
```
dartcafe: The type STRING was by intention and a preparation for another feature.

tortuetorche:
OK, but what's the default value for 'show_result'? Because it can't be null in the database.

tortuetorche:
> 
> 
> 'notnull' => true was by intention.

SQLite can't have a NOT NULL constraint without DEFAULT value.

I understand the need to require a value for the `delete_date` column, but it should be done on the application level to support SQLite.

FYI, Nextcloud Server doesn't have any DATETIME column with a NOT NULL constraint, see:
https://github.com/nextcloud/server/blob/v17.0.1/core/Migrations/Version13000Date20170718121200.php#L752

Here the SQLite error message:
```
An exception occurred while executing 'ALTER TABLE oc_polls_events ADD COLUMN delete_date DATETIME NOT NULL':

  SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL
```

dartcafe:
> 
> 
> > 'notnull' => true was by intention.
> 
> FYI, Nextcloud Server doesn't have any DATETIME column with a NOT NULL constraint, see:
> https://github.com/nextcloud/server/blob/v17.0.1/core/Migrations/Version13000Date20170718121200.php#L752

I was focused on MySQL, and did not respect other databases. Have to make a note to myself here... 

Anyway, great support saved a lot of time for me.

tortuetorche:
Hi @dartcafe,

Thank you for merging this 👍 
The Nextcloud Polls application is great and I'm very enthusiastic about the upcoming 1.0 version!

Have a good day,
Tortue Torche

Issue #723, opened 2020-01-09
1.0b3: comments throw fatal errors in log


herrmannsdorfer:
As soon as I comment on a poll the log is filled with "fatals". Maybe related with notifications (#667)?

![Bildschirmfoto 2020-01-09 um 15 00 17](https://user-images.githubusercontent.com/43957716/72073720-c4972100-32f0-11ea-8388-106ebc442a30.png)

Issue #722, opened 2020-01-09
1.0b3: some minor details in the comments section


herrmannsdorfer:
- The contents of your last comment are not cleared from the entry field.
- It would be helpful to post your comment by just hitting enter instead of mouse movement and clicking.

![Bildschirmfoto 2020-01-09 um 12 55 07](https://user-images.githubusercontent.com/43957716/72066056-f18f0800-32df-11ea-9348-8a16cb7b513f.png)

Issue #718, opened 2020-01-08
1.0b2: "Delete" is missing


herrmannsdorfer:
- The delete function does not work.
- The clone function does not work either or I do not understand how to use it.

herrmannsdorfer:
The "Delete poll" button in the config pane works. The option in the overview (...) does not.

herrmannsdorfer:
When a poll ist already deleted, the option to delete it should be omitted in the three dot menu in the overview (or replaced with "restore poll" instead).

herrmannsdorfer:
As deleted polls are also listed in the "All polls" overview (should they?) they should have an "deleted" icon to distinguish them from open polls. Red icon perhaps?

dartcafe:
poll management will return later. I removed the options for the moment, as they do not work as expected.

dartcafe:
When deleting a poll, it diapperas from all lists and appears in the deleted polls on my test site.

herrmannsdorfer:
Just tried with a new user and fresh polls. Deleted two of them and added this info to the title. The deleted polls are not in "All polls" in the left pane but still listed in the main view. They are of course listed in "Deleted polls".

![Bildschirmfoto 2020-01-09 um 14 36 25](https://user-images.githubusercontent.com/43957716/72072427-33bf4600-32ee-11ea-82b8-9420f0557efa.png)

dartcafe:
Ah. I only watched the navigation. Will check that.

PR #724, opened 2020-01-09
Fix comments

dartcafe:
#723 and #722
#718

### DIFF
--- a/src/js/components/Comments/CommentAdd.vue
+++ b/src/js/components/Comments/CommentAdd.vue
@@ -27,8 +27,8 @@
 		<form class="commentAdd" name="send-comment">
 			<input v-model="comment" class="message" data-placeholder="New Comment ..."
 				@keyup.enter="writeComment">
-			<input v-show="!isLoading" class="submit-comment icon-confirm"
-				@click="writeComment">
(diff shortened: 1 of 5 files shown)


src/js/components/Comments/CommentAdd.vue
```diff
@@ -25,8 +25,10 @@
 		<user-div :user-id="currentUser" />
 
 		<form class="commentAdd" name="send-comment">
-			<input v-model="comment" class="message" data-placeholder="New Comment ...">
-			<input v-show="!isLoading" class="submit-comment icon-confirm" @click="writeComment">
+			<input v-model="comment" class="message" data-placeholder="New Comment ..."
+				@keyup.enter="writeComment">
+			<input v-show="!isLoading" class="submit-comment icon-confirm"
+				@click="writeComment">
```
v1r0x: Is there a reason you are using an input field as button?

If not, you could do something like this:
```vue
<form class="..." name="..." @submit="writeComment">
    <input v-model="comment" class="message" data-placeholder="New Comment ..." />
    <button type="submit" v-show="!isLoading" class="icon-confirm">...</button>
</form>
```

PR #747, opened 2020-01-15
Compare date poll with user's calendar

dartcafe:
First attempt
grafik

grafik


dartcafe:
Currently blocked by https://github.com/nextcloud/server/pull/18679

jancborchardt:
Very nice! :)
- [ ] Let’s use the calendar icon instead of an error icon
- [ ] Icon can be white instead of black on the orange color
- [ ] If we know the time of the poll option, we should list the conflicting event directly, and only the relevant one. So right of the icon:

> 10:00 – 11:00
> Design team meeting

dartcafe:
>     * If we know the time of the poll option, we should list the conflicting event directly, and only the relevant one. So right of the icon:
> 
> 
> > 10:00 – 11:00
> > Design team meeting

I decided to put it into the tooltip, because we could have a situation like this: 
grafik

There will be no space to add this additional information, IMHO.

jancborchardt:
If we show it in the row where you can reply, it might make a bit more sense than in the top bar?
Then instead of the icon, we can say "Calendar conflict" or something short like that which is more understandable.

dartcafe:
> we can say "Calendar conflict" or something short like that which is more understandable.

Hmm. I would prefer the andon way. 😄

v1r0x:
Great feature! Looks already very nice!

I'd also move it to the users row, so it's better visible while going through the options. I have no real opinion whether using tooltip or not, but I'd suggest focusing on the important stuff and only show the name of the event (maybe from which calendar?) and a "... and 3 more events", if there are several conflicting events. Without a tooltip we could add more info to all of these events to a tooltip.

Here's a screenshot from a version without tooltip and icon moved down to the users row

polls-calendar

dartcafe:
In my optinion the text will break the design, if there are a lot of options, so I would keep the tooltip. Regarding moving the icon, I will try some variants in a few weeks. Nevertheless, we are dependent on the release of the PR in the server repo.

dartcafe:
And BTW: Any idea, how to create a calendar event, without building a dav client in polls? I looked at the implementation in the mail app. Not what I expected.

dartcafe:
@georgehrke @tcitworld  Is there a simple way to add an event to the calendar, without building a DAV client, like it seems that is done in the mail app?

tcitworld:
Managing calendars can be done with `OCP\Calendar` but for managing events I'm afraid you currently need to hook in directly into `OCA\DAV\CalDAV\CalDavBackend`.

dartcafe:
@tcitworld any walkthru or pattern available?

> but for managing events I'm afraid you currently need to hook in directly into `OCA\DAV\CalDAV\CalDavBackend`.

tcitworld:
What you want is just calling `CalDavBackend::createCalendarObject($calendarId, $objectUri, $calendarData)` with iCalendar data for $calendarData. See the tests for instance: https://github.com/nextcloud/server/blob/master/apps/dav/tests/unit/CalDAV/CalDavBackendTest.php#L206-L232

dartcafe:
@tcitworld  Thanks. I think that helps me getting into it.

jancborchardt:
> In my opinion the text will break the design, if there are a lot of options, so I would keep the tooltip. Regarding moving the icon, I will try some variants in a few weeks. Nevertheless, we are dependent on the release of the PR in the server repo.

Since the info about the conflict is the most important thing here, and tooltips are not very discoverable, it should be directly shown. Even just "Conflicts with **Event name**" would be great, the calendar icon on orange is not really needed cause the text already says it.

dartcafe:
> Since the info about the conflict is the most important thing here, and tooltips are not very discoverable, it should be directly shown. Even just "Conflicts with **Event name**" would be great, the calendar icon on orange is not really needed cause the text already says it.

I would say, at first sight it is not important to know which calendar entry causes the conflict. In fact the time before and after the poll date is interesting as well, because, you have to think about travel times or similar.

dartcafe:
I will try some mockups to illustrate the problem I see.

dartcafe:
Closing this, as the PR is rather old. I ported the attempt to #1056

Issue #717, opened 2020-01-08
Sort polls in poll list


herrmannsdorfer:
- There should be a way to sort polls in the left overview "All polls" or in the main view, eg by dragging an item or selecting a column header.
- I would prefer a standard sorting by youngest at the top. This way you don't have to scroll down to see the latest entries (new polls or polls with recent votes or comments).
- Maybe it would be a good idea to pin selected polls as favorites at the very top of the list.

dartcafe:
Sorting is still missing and will come with a better poll management.
Favorites are not planned for now. there are things of more worth. You can open an issue for that.

dartcafe:
#586

Issue #594, opened 2019-05-31
Drop down menus not clickable in mobile view


coxtor:
### What is going wrong?
**Describe the bug**
When viewing the app on a mobile browser the drop down sections are unclickable. This also applies when using the "view desktop version" functionality of the browser. The issue persists on at least Firefox and Chrome mobile. 

**To Reproduce**
1. Open the app and create a poll
2. Return to poll overview
3. Click on the three dots to open the dropdown
4. Try to select any element

**Expected behavior**
The menu points are clickable.

**Screenshots**

### Information about your polls installation
0.10.2

**Fresh installation or update from a prior version (from which one)?**
docker nextcloud 16, issue also occurs on nextcloud 15

**How did you install this version?(Appstore or describe installation)**
appstore

welshcoder:
Hi,

I can confirm this bug also appears in Safari using iPhone iOS 11.4 and in Firefox 68.0 on Android 8.0.0 on a Samsung Galaxy S7 too.
Using Polls 0.10.2, NC 16.0.3 (manual clean install) and PHP 7.1 on FreeBSD 11.2.
I hope that helps.

Ich5003:
Bug is still present.
Chrome on Android 10, Pixel 3a

In my opinion this is one of the most important bugs to fix, because not being able to share a poll on mobile phone prevents using this in any productive area

Issue #559, opened 2019-03-28
New polls should appear at the top


ElkbA:
It would be great if one could sort the polls so that the newest polls would appear at the top of the page.

dartcafe:
After finishing the vue migration, the navigation is planned to give an eye to the polls, which need attention of the user.

PR #753, opened 2020-01-20
Sort polls in poll list

dartcafe:
* changing popovermenu to Actions (#594 and #754 )
* updated design according to file lists
* Sorting polls in poll list, default creation date desc (#559 and #717) 

grafik
grafik
grafik


dartcafe:
I'll merge this, because we need a 1.1 shortly.

Issue #761, opened 2020-01-21
Existing polls not available after update to polls 1.0.0


domchrgit:
After update to nextcloud/polls 1.0.0 (from version 0.10.4) my existing polls are not visible.
If I browse the MySQL database, the entries are already there, but nothing is available in the new nexcloud/polls

How to recover the existing polls ?

Nextcloud 17.0.2 / Polls 1.0.0 / PHP7.3

Art4:
I also have this issue. The problem is that the sharing url has changed from `https://example.com/index.php/apps/polls/**poll**/abc` to `https://example.com/index.php/apps/polls/**s**/abc`.

It would be nice if there would be a redirection from the old urls to the new ones.

dartcafe:
@domchrgit Can you explain a little bit more? 
* What do you see (screenshot)?
* a sample of your db entries would help (oc_polls)

@Art4 I added a redirection from /poll/* to /s/* for public polls.

domchrgit:
> @domchrgit Can you explain a little bit more?
> 
> * What do you see (screenshot)?
I see an empty page in polls tab since I had a only a single public poll made with previous version...
But effectively, outside of nextcloud, if I change the link as suggested by @Art4, I manage to view my old public poll.
Perhaps with the redirection you add, I will see again the poll in nextcloud after updating with the next release of polls app ?

> * a sample of your db entries would help (oc_polls)
> 
> @Art4 I added a redirection from /poll/* to /s/* for public polls.

dartcafe:
@domchrgit Ah. OK thanks for explaining. 
It depends on the settings in the polls and who created the poll. If another user created the poll and it is set to hidden, only invited users can see the poll and the creator.

domchrgit:
@dartcafe The poll was created by me, as administrator , and it was shared as a public poll...
Under my session I don't see the poll now with version 1.0.0. and the "old" link doesn't work without the modification suggested by @Art4 .
When do you expect to release the new version with your correction in Nexcloud/Polls ?

Art4:
@domchrgit This bug will be fixed with 923d0f106c92e8699b966ea1721abfe15cd8824f.

@dartcafe The problem might be that there is no 404 page or redirect to login. You can reproduce this with an invalid share url like https://example.com/index.php/apps/polls/s/a This shows a white page and logs an error in the console.

dartcafe:
Screenshots and a dastabase dump of the poll would help to understand.

dartcafe:
[...]

Issue #759, opened 2020-01-21
Starting a poll automatically says you have participated


DBJRdev:
### What is going wrong?
**Describe the bug**
In Polls v1.0 the creator of a poll is automatically shown as attendee, right from the start. Once one or more vote options are defined, Polls says the creator has voted "no" in the "bars" view (red bar with "1" is visible, see screenshot below). In closed bug report #556 it was said that the creator would not be counted unless s/he has participated actively, but apparently the behaviour of the polls app has changed since then.

It is very confusing for users, even if you know the above-described behaviour – because you will then always not trust the numbers, but start counting by hand.

**To Reproduce**
1. Go to polls app, create a new poll
2. Create the first options
3. Go to "change view" ("Ansicht wechseln") and see that you're counted without even having attended

**Expected behavior**
A creator should not be counted as participant, unless s/he has really cast the votes.

**Screenshots**
The screenshot shows what appears automatically. The red arrow and question mark was added by me.
![20200121 polls v1 creator appears to have voted](https://user-images.githubusercontent.com/13522599/72766871-0ded4780-3bf2-11ea-9e15-a4107e75be16.png)

### Information about your polls installation
**Polls version?** 
v1.0

**Fresh installation or update from a prior version (from which one)?**
Updated via app store

### Information about your Instance of Nextcloud/ownCloud
**Nextcloud or ownCloud?**
NC 17.0.2

Issue #757, opened 2020-01-20
When creating a new poll, the cursor should automatically be inside the polls name field


szaimen:
**Is your feature request related to a problem? Please describe.**
When creating a new poll, the cursor should automatically be inside the polls name field. Then the polls creating flow would be much better, instead of manually have to click in this field

**Screenshot**
This is directly after clicking on the new poll field (the cursor isn't automatically inside the title field)
![image](https://user-images.githubusercontent.com/42591237/72761350-0c188980-3bdc-11ea-8de0-6f015a5b846b.png)

szaimen:
@dartcafe is this done?

PR #762, opened 2020-01-21
Fixes 1.0

dartcafe:
* autoFocus poll title in creation dialog (#757)
* only count current user, if he actually voted (#759)
* redirect old public polls routes to new route (#761)
* Avoid error on creating share (#763)


dartcafe:
@v1r0x I'll merge this too. We need a 1.1 fter that. Will do some tests after that

PR #781, opened 2020-01-25
Save username when enter is pressed on public link

SystemKeeper:
Currently it's only possible to save a username when accessing a public link with the arrow-button. This PR allows pressing "enter" as well.


dartcafe:
Already part of https://github.com/nextcloud/polls/tree/modalUsername
PR #784

Issue #773, opened 2020-01-23
Deleted polls shouldn't be accessible through a public link


SystemKeeper:
### What is going wrong?
**Describe the bug**
Not sure if this is intended behaviour or not. When a poll is deleted, it is moved to "Deleted polls", which is fine. Personally I would've expected that public links will be kept (in case a poll is restored afterwards), but wouldn't be accessible through this public link.

**To Reproduce**
1. Create a poll
2. Create a public link
3. Delete the poll
4. Try to access the poll with the created public link.

**Expected behavior**
As the link is a public link, I would've expected to get an error (same as you were trying to access a non-existant public link), but instead the poll is still shown. 
Also: I can still enter a Name and create a new personal link! I cant vote for anything but I could add public links.

**Screenshots**
![image](https://user-images.githubusercontent.com/1580193/73033020-0ec7e880-3e41-11ea-874f-f62100344bff.png)

### Information about your polls installation
**Polls version?** 1.1.5

**Fresh installation or update from a prior version (from which one)?**

Update

**How did you install this version?(Appstore or describe installation)**
Manual

### Information about your Instance of Nextcloud/ownCloud
**Nextcloud or ownCloud?** NC

**Which Version?** 17.02

SystemKeeper:
After thinking about this: I still believe a deleted poll shouldn't be accessible by anyone else than the owner, but maybe we could add the possibility to switch a poll to "read-only" state. This way the results would still be visible to all participants (as well as through the public link), but no changes would be allowed.

Issue #772, opened 2020-01-23
Accessing a non-existant share shouldn't fail silently


SystemKeeper:
### What is going wrong?
**Describe the bug**
When accessing a share which is not available anymore a white page with nextcloud header is displayed instead of an error.

**To Reproduce**
1. Create a poll
2. Create a public link
3. Remove the public link
4. Try to access the now removed link

**Expected behavior**

**Screenshots**
Show a error message like nextcloud does when accessing a non-existant file-share:
![image](https://user-images.githubusercontent.com/1580193/73032312-1c7c6e80-3e3f-11ea-94ef-ec76c3f37866.png)

Current behaviour:
![image](https://user-images.githubusercontent.com/1580193/73032344-29995d80-3e3f-11ea-8a9b-fe5301c457d6.png)

The ajax request does already fail with 404:
![image](https://user-images.githubusercontent.com/1580193/73032426-5baabf80-3e3f-11ea-9256-87b469606f65.png)

### Information about your polls installation
**Polls version?** 1.1.5

**Fresh installation or update from a prior version (from which one)?**
Update

**How did you install this version?(Appstore or describe installation)**
Manually

### Information about your Instance of Nextcloud/ownCloud
**Nextcloud or ownCloud?** NC

**Which Version?** 17.0.2

PR #784, opened 2020-01-26
Modal username

dartcafe:


SystemKeeper:
@dartcafe This interferes with #772 and #773 
image

The redirect is working correctly (`/not-found`), but the modal shouldn't be there.

Issue #715, opened 2020-01-08
Provide displayNames instead of userId


herrmannsdorfer:
#402 is back.

![Bildschirmfoto 2020-01-08 um 13 42 17](https://user-images.githubusercontent.com/43957716/71979074-1a9a9480-321d-11ea-922d-3842a3d7f966.png)

herrmannsdorfer:
Looking at the screenshot I would also take care of the column width (created entries overlapping expiration date) and the vertical alignment of the owner to the other text lines.

dartcafe:
I have to fiddle out, how to get the displaynames via the js API. comes later.

tortuetorche:
Hi @dartcafe,

First of all, thank you for this great Nextcloud App 👍 

Here some clues to display names instead of user id, based on [Nextcloud Talk/Spreed](https://github.com/nextcloud/spreed) code :
Nextcloud Talk use the [`@nextcloud/auth`](https://github.com/nextcloud/nextcloud-auth) NPM module, here the reference: https://github.com/nextcloud/spreed/blob/v8.0.0/package.json#L18

Then, here the code to load this module: https://github.com/nextcloud/spreed/blob/v8.0.0/src/PublicShareAuthSidebar.vue#L37

Here the code to manipulate the `getCurrentUser()` object: https://github.com/nextcloud/spreed/blob/v8.0.0/src/PublicShareAuthSidebar.vue#L164

And finally, the code to extract the `displayName` of the user, if any: https://github.com/nextcloud/spreed/blob/v8.0.0/src/store/actorStore.js#L135

Have a good day,
Tortue Torche

tortuetorche:
Hi folks,

I think my [previous comment](https://github.com/nextcloud/polls/issues/715#issuecomment-578190370) go in the wrong direction, because it's only focused on the current user 😕 

Cheers,
Tortue Torche

dartcafe:
Yes, right. The displayNames are AFAIK not available via the JS API. But I want to avoid loading every username via a individual API-Call. That means, I have to load them when gathering the information in the backend. This means extending the models. I need some time for that. Stabilizing the app has a higher priority ATM.

Dennis1993:
I have the same problem, I don’t know who has voted 😅🤷‍♂️
NC 17.0.2 on Ubuntu with Polls 1.1.6

dartcafe:
See beta4 in the releases. There are some more enhancements and fixes. please report bugs in new issues.

[...]

PR #798, opened 2020-01-29
Add display names

dartcafe:
Show displayName instead of userId #715

### DIFF
--- a/lib/Controller/CommentController.php
+++ b/lib/Controller/CommentController.php
@@ -142,11 +142,14 @@ class CommentController extends Controller {
 	 * @NoAdminRequired
 	 * @NoCSRFRequired
 	 * @param int $pollId
+	 * @param string $userId
 	 * @param string $message
 	 * @return DataResponse
 	 */
 	public function write($pollId, $userId, $message) {
+		$this->logger->alert('write');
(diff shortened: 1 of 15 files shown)


herrmannsdorfer:
`displayName` is now shown everywhere except in the header:

Screenshot 2020-01-29 at 16 58 16

And the log reads:

Screenshot 2020-01-29 at 17 03 28

Notification is not sent.

dartcafe:
Had some trouble with the userBubble component. Is fixed. I'll wait for some bugs to fix before makint the final RC of Polls 1.2

dartcafe:
Notification are working on my side. There was an error in the route. Maybe, the job is paused. Normally it will resume automatically after timeout. Otherwise look in the OC_Jobs table and set all timestamps (`last_run` and `last_checked`) of the job `OCA\Polls\Cron\NotificationCron` to zero.
If there is an entry under `reserved_at`, the cron is paused because of errors. Set it to zero and it will be queued again.

herrmannsdorfer:
Confirmed. Notifications work. But ...

Activity notification still uses userId instead of displayName :-)
Screenshot 2020-01-30 at 09 35 40

dartcafe:
Missed that ones.

PR #823, opened 2020-02-07
Delete poll permanently

### LINKED ISSUE #801: Delete a poll completly
It is not possible to delete a poll permanently.
If we do some polls the list will be long :D

image

v1r0x:
fix #801


tortuetorche:
Hi @v1r0x,

I think, you should rebase this branch against `master` branch to pass tests

Have a good day,
Tortue Torche

dartcafe:
Tavis can be ignored, because it bases on a known error, which is fixed in the master branch. In the end there are no valuable tests...

appinfo/routes.php
```diff
@@ -55,6 +55,7 @@
 		['name' => 'poll#list', 'url' => '/polls/list/', 'verb' => 'GET'],
 		['name' => 'poll#get', 'url' => '/polls/get/{pollId}', 'verb' => 'GET'],
 		['name' => 'poll#delete', 'url' => '/polls/delete/{pollId}', 'verb' => 'GET'],
+		['name' => 'poll#deletePermanently', 'url' => '/polls/delete/permanent/{pollId}', 'verb' => 'GET'],
```
splitt3r: Any reason for not using the `DELETE` verb?
```suggestion
		['name' => 'poll#deletePermanently', 'url' => '/polls/delete/permanent/{pollId}', 'verb' => 'DELETE'],
```

v1r0x: Itried it, but it didn't work. Thus I thought this isn't supported in NC

splitt3r: I think it should 😁 a quick research gave me:

https://github.com/nextcloud/spreed/blob/master/appinfo/routes.php#L92-L100
```php
...
[
		'name' => 'Call#leaveCall',
		'url' => '/api/{apiVersion}/call/{token}',
		'verb' => 'DELETE',
		...
]
...
```
https://github.com/nextcloud/spreed/blob/master/src/services/participantsService.js#L75

```javascript
...
const leaveConversationSync = function(token) {
	axios.delete(generateOcsUrl('apps/spreed/api/v1/room', 2) + token + '/participants/active')
}
...
```

v1r0x: For some strange reason it worked fine this time 😕 Maybe I did something wrong last time. Will push a fix when at my dev machine.

dartcafe:
Routes are still a little bit unsorted and could need some optimization. More verb using.

appinfo/routes.php
dartcafe: FYI. Building works well on windows machines with some tools. Only signing is an issue.

dartcafe:
@v1r0x Any further to do here?

dartcafe:
I solved the merge conflict and changed to ButtonDiv.
After pull, make an `npm install` to resolve the dependencies merged from master!

v1r0x:
Changing verb from get to delete is still not done. Beside that I'm ready to merge. Will fix on monday.

dartcafe:
> Changing verb from get to delete is still not done

Not really importend for now, IMHO. It works the way, you resolved it. As the routes need some refactoring, this could be done later, together with the rest and maybe by switching to ressources in the routes.php.

v1r0x:
Then lgtm :)

PR #842, opened 2020-02-19
Dependency and config tidy

dartcafe:
@v1r0x I did some clean up to the installed packages and removed some. Could you review the changes? It seems to me, that some dependencies were installed, but they were not used (especially some babel packages). 

Added terser-webpack-plugin for (slightly) smaller files and removing comments from prod  JS

# Changes: 
## Removed unused dependencies
* @babel/runtime
* acorn
* acorn-jsx
* babel-plugin-transform-async-to-generator
* babel-plugin-transform-object-rest-spread
* babel-plugin-transform-runtime
* babel-preset-latest
* fibers
* v-tooltip
 
## removed unused devDependencies
* @babel/plugin-syntax-dynamic-import
* babel-preset-env
* prettier-eslint
* raw-loader
* eslint-plugin-nextcloud

## Configuration to be changed
.eslintrc.js
'comma-dangle': ['error', 'always-multiline'] 
--> makes editing easier, no more errors on trailing commas, `npm run lint:fix` will make the changes

What do you think?


dartcafe:
@v1r0x Added Babel again and use @nextcloud/eslint 
Configuration seems to work, so I would like to merge this to the master and do some rebasing. 

Any opinion of using 'comma-dangle': ['error', 'always-multiline'] for eslint?

dartcafe:
> but `'always-multiline'` is fine.

Because this will affect all branches, I suggest to introduce this after the next release.

PR #845, opened 2020-02-22
workaround darkReader (browser add-on)

dartcafe:
fix text color change in dark mode, when using darkreader add-on (https://github.com/darkreader/darkreader)


dartcafe:
Reviewed by myself. ;-)

PR #854, opened 2020-02-29
Remove users from a poll

### LINKED ISSUE #736: Remove Participants from vote
Poll owner should be able to remove participants and their votes from a poll.

dartcafe:
fix #736


dartcafe:
> Looks good!

Not a pretty solution IMHO, but OK for now. Maybe someone has a better UX idea.

dartcafe:
> Looks good!

Not a pretty solution IMHO, but OK for now. Maybe someone has a better UX idea.

PR #855, opened 2020-03-01
create date option sequence

### LINKED ISSUE #649: Create series of options
Thanks for this nice app!

I couldn't find an easy way to create many time slot options at once. Creating each option
individually is too time consuming in some instances.

**A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]**

I need to find a 30 minutes slot for 10 people sometime between 22.10 - 25.10. Only slots
between 2pm - 5pm are possible.

After selecting a date I would like to be able to select multiple times 2pm 3pm etc that are added as separate options to the poll. Being able to repeat this for other days easily (like the shift option, but copy to other day instead) would be even better.

**Describe alternatives you've considered**
---

**Additional context**

 Selecting multiple times #326

dartcafe:
fixes #649


dartcafe:
First commit:
grafik
grafik
grafik

tortuetorche:
Hi @dartecafe,

This new feature looks cool  😎, but IMHO now deleting an option is less intuitive/ergonomic for new/beginner users.
Because they need to click on the three dots icon to found the delete action.
Is it possible to show the delete icon directly and add the three dots icon on its right side?

Have a good day,
Tortue Torche

dartcafe:
@tortuetorche Aggree. It is also annoying, when trying to delete a couple of options. I separated the actions into two.

dartcafe:
@v1r0x any comments or ready to merge?

v1r0x:
I've checked a older version and that looked fine. I'm kinda busy this weekend. Next week I can do some final checks.

PR #889, opened 2020-04-07
[1.4.0 beta1] Fix datetimepicker

### LINKED ISSUE #826: Allow all time-possibilities for date/time-polls
**Which Polls version are you running?** (see apps page)
Polls Vers. 1.1.6 and Nextcloud Vers. 17.0.2

### Steps to reproduce
Create a new poll.
Add a new Date-option

### Actual behaviour
You can only select time at the full hour or at .30. Other times can't be selected.

### Expected behaviour
I should be able to select whatever time I want to.

As the trains for the participants of our meetings arrive at the full hour and they have to walk about 15 minutes to the office, the meetings normaly start at a quarter past the hour. If I don't state a clear time in the poll, the result will not be correct, as people think, they have to take a train an hour earlier.

dartcafe:
* fixed setting of expiration date 
* updated @nextcloud/vue to latest version (1.4.1) at least for security fixes
* updated date-picker options to new format (vue2-datepicker 3.4.1, which came with @nextcloud/vue 1.4.1)
  * switched to `minuteStep: 1` fixes #826
* Changed picker layout 
grafik

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
 	"dependencies": {
 		"@nextcloud/axios": "^1.3.1",
 		"@nextcloud/router": "^1.0.0",
-		"@nextcloud/vue": "^1.3.0",
+		"@nextcloud/vue": "^1.4.1",
 		"core-js": "^3.6.4",
 		"lodash": "^4.17.15",
 		"moment": "^2.23.0",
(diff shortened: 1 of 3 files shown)


dartcafe:
Merged in favour of a new beta with some fixes.

PR #890, opened 2020-04-10
set window title

### LINKED ISSUE #318: Poll Title should also go into the html-document's `<title>` tag
### Current behaviour
When I navigate to a poll, the browser title will remain the default "Umfragen" ("Polls") for every single poll. This is a problem for navigating one's history or also for identifying polls from the browser tab title.

### Expected behaviour
The Poll's title should be part of the document title, so that identification is easier.

To Dos
- [X] Create/Edit page
- [x] Vote page

dartcafe:
fixes  #318

### DIFF
--- a/src/js/views/PollList.vue
+++ b/src/js/views/PollList.vue
@@ -119,6 +119,10 @@ export default {
 			}
 		},
 
+		windowTitle() {
+			return t('polls', 'Polls') + ' - ' + this.title
+		},
+
 		sortedList() {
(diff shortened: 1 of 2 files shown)


dartcafe:
Merging to get this into beta2

ccoenen:
I'm not sure if this was lost or something, but the polls don't have a title for me today. Neither as a logged in editor of a poll, nor as a visitor of a shared poll.

PR #893, opened 2020-04-10
hide results

### LINKED ISSUE #265: Hide results while poll is open
An option to make a poll completely anonymous by not showing the choices made by other users.

dartcafe:
resolves #265 
Added "Show results" to configuration:
* always
* after poll expiration
* never

grafik

The poll owner will see the results always and the participants get an information, that the result is hidden.
grafik


dartcafe:
Merging , because this changes accidetially are occuring in #896

PR #897, opened 2020-04-10
Add poll description to invitation mail

### LINKED ISSUE #895: 1.4.0 Beta: Description missing in invitation email
It would be nice to have also the poll description in the invitation email in case a description is given in the poll.

image

... should go here:

image

dartcafe:
resolves #895


dartcafe:
simle addition, so merging after test without review.

PR #922, opened 2020-05-08
Vue refactoring

dartcafe:
some refactoring of the complete vue app and initial setup for 1.5
❗  Please rebase your current branches after this is merged to master

* fixing of style issues
* got rid of some terrible CSS constructions
* moved colors and icons to root variables

* merged some components to avoid double code
* cleanup store (some more consolidation is a to do)

* updated to vue-nextcloud 2.0
* using more @nextcloud-modules 
* updated dev dependencies
* updated linter config
* updated polls version to 1.5.0

* load a poll in one request
* make sidebar available in poll list
* added in-/decremet buttons for moving date options

PR #929, opened 2020-05-12
Confirm options

dartcafe:
Merge after #922 

Mark options as confirmed after the poll expired
grafik
grafik


dartcafe:
@v1r0x You can test this PR as well, it is based on #922

dartcafe:
I added a max-width. Not happy with that too, but design fixes can be done in smaller PRs.

dartcafe:
>     1. mobile view you can not confirm/unconfirm an option in the main view (or I did not find it 😅 )

In mobile mode, you have to use the sidebar. In the end, this ias also a question of design/UX as the mobile view in general. There is room for improvement.

dartcafe:
> In desktop view the bottom part of the confirm border is cut off, but not in mobile view (**only for non-admin users!**)

👍 , I'll fix this

dartcafe:
@v1r0x Did you check the fix? If we merge this and #942, we could continue with smaller PRs.

PR #939, opened 2020-05-27
graphics, photos to vote for

delete-exe:
Hi,
at the moment there are only dates and text options to vote for in a poll.

Is it possible that you add graphics, photo poll option, so that you could start a voting for different photos or i.e. layouts that your team members should make a decision about? You should see at the end with layout got the most clicks.

My best regards


dartcafe:
@delete-exe You created a PR. Please open an issue.

PR #942, opened 2020-05-28
Namespaced store

dartcafe:
Changed store modules to submodules and use namespacing.


dartcafe:
@v1r0x this PR is based on #929

v1r0x:
Can you give me some more insight on what this PR should add/fix?

dartcafe:
@v1r0x pure refactoring

Before this we had several modules in the vue store on the same level. 
Now: 
* Votes, options, etc. got submodules of polls
* all modules got namespaced, so that they have to be called explicitly with their namespace

For example:  
* `.dispatch('addOption', ...)` got changed to `.dispatch('poll/options/add', ...)`
* `store.options` got `store.poll.options`
* `.commit('set')` called all set mutations in all modules simultaneously, now they are called individually `.'commit('poll/set')`, `.commit('poll/options/set')`

Additionally 
* I had to disable the CSRF preconditions temporarily, because there is something strange with the token generation. I get CSRF errors after some time. Anyhow the polling for new tokens is broken. This has to be inspected and fixed seperately. Maybe, there is some misbehavior in the current RC of NC19 (my dev/test instance is on the beta channel).
* reordering of the text polls got moved to the backend to have it behind future API calls.

dartcafe:
You can see the differrences best with the vue Devtools addon
before: 
grafik

after: 
grafik

dartcafe:
late deliver - renamed lists in store to propertyname 'list' for clarification

PR #954, opened 2020-06-11
avoid fetching foto

### LINKED ISSUE #912: slow ldap user search
Hello, 

when searching Users to invite, it takes very long to find a user. (up to 10 min) 
The user search in the file sharing app works very fast. Any tip on what I could check ? We are connected to a AD with several thousand users.. but as I said only in the polls invitation it is very slow. 

thank you.

dartcafe:
fix #912


dartcafe:
Merging via #1038

PR #955, opened 2020-06-11
Refactoring vote clicks

### LINKED ISSUE #920: Change click order when "Maybe" is enabled
From my experience users who do not know the app in detail or are first-time users (mostly external ones), don't use it correctly when the option "Maybe" is enabled.

The click order gives you "Maybe" with the first click and "Yes" with the second click - 3 out of 3 external users in a poll just clicked the dates they are available and they were all set to "Maybe". I needed to tell them that they need to klick again.

In my opinion it would be better to have "Yes" as the first click choice and "Maybe" with the second click.

dartcafe:
* fix #920 
* moved vote sequence from votes store to poll store
* handle vote click inside of vote item (VoteTableVoteItem.vue)

### DIFF
--- a/src/js/components/VoteTable/VoteTable.vue
+++ b/src/js/components/VoteTable/VoteTable.vue
@@ -52,8 +52,7 @@
 					:key="option.id"
 					:user-id="participant.userId"
 					:option="option"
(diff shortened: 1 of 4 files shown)


dartcafe:
Merging via #1038

PR #1016, opened 2020-07-11
Tidy and doc

dartcafe:
* Mainly code maintenance
  * removed unused functions
  * renamed controller routes
  * added, deleted and updated annotations
  * added and corrected type declarations
  * updated ajax calls from store due to changes in the controller routes


dartcafe:
@v1r0x I tested a lot while and after working on the files. I think the controllers are fine now and the annotations are ok. Maybe I missed some annotations. Also the REST-API (pretty good testable with postman) should work and the docs should be up to date.

dartcafe:
merging via #1038

PR #1105, opened 2020-09-17
Default view for poll types

### LINKED ISSUE #744: Preference for default view
In polls 1.0 the default view changed from "doodle basic" table layout to a new list view. A "Switch view" button was added to change the mode. Although the last state was not remembered when you log off.

The new list view is a valuable addition. It gives a condensed, analytical overview.

The old layout is more detailed and probably familiar to anyone with doodle experience.

- I would suggest that the last state of the view is remembered. Moderators might want to use the new list view, other participants may prefer the "classic" view.
- To mitigate the layout change I would also suggest that the default view remains the table layout. That will make it easier for new poll participants who doodled before.

dartcafe:
fix #744 

Define the default view for text and date polls
* The initial default setting for text polls is now mobile view (list layout)
* The initial default setting for date polls is now desktop view (table layout)
* The user can define his preffered setting
* Changing the view mode is possible und is kept for the poll type, when switching between polls
* initial setting on mobile layout is mobile view, regardless what default is set

### DIFF
--- a/src/js/components/Navigation/NavigationSettings.vue
+++ b/src/js/components/Navigation/NavigationSettings.vue
@@ -26,6 +26,14 @@
 			type="checkbox" class="checkbox">
 		<label for="calendarPeek">{{ t('polls', 'Use calendar lookup') }}</label>
 
+		<input id="defaultViewTextPoll" v-model="defaultViewTextPoll"
+			type="checkbox" class="checkbox">
+		<label for="defaultViewTextPoll">{{ t('polls', 'Text polls default to table view') }}</label>
+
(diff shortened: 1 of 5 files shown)


dartcafe:
@v1r0x Merging to head to another RC.

herrmannsdorfer:
Just checked 1.5RC3: Mobile view of a date poll is different now. In previous releases the bar length represented the number of votes. Now it is always 100 % and only the the yes (green) and maybe (orange) votes are stacked proportionally.
Readability is bad, previous versions were easier to interpret.
Result is browser or device independent.

dartcafe:
Thanks, indeed a bug. But would be better to create a new issue, to get aware of it.

PR #1112, opened 2020-09-20
public polls - new dialog for login

### LINKED ISSUE #1096: Bigger Sign In Option
A lot of Cloud users follow the provided link to the poll, enter a username and thus particpate without their account.
That creates a few issues:
Users complain, that their name ist already in use.
Users assume they can change the poll via their account, which doesn't work.

An option to prioritise Sign in over participating without account would be great. Perfect would basically be to change the current options. 
Big the link to sign in, and a small link to participate without account.

Thx

dartcafe:
fix #1096 
public polls - Added dialog to ask if user has login or not


dartcafe:
@v1r0x merging too

PR #1155, opened 2020-10-13
fix date shifting

### LINKED ISSUE #1154: Shift all dates changes times to UTC
Goal -- set up a list of options across days send out poll, then clone and update for next week maintaining time zone.

Problem -- if you set up a poll using time zone of "America/New_York" and then use "shift all dates" to try and accomplish this, all of the resulting time slots are in UTC.

dartcafe:
fix #1154


dartcafe:
checking in the bugfix without review

PR #1219, opened 2020-11-04
Error handling

dartcafe:
* use own exceptions class
* changed error handling
* fixed - public and hidden values
* reduced complexity of acl class
* js - removed  concat methods
* added traits for conroller responses to make sure, responses are consistent for all controllers and avoid duplicated code
* added fix step to migration to remove invalid settings entries (duplicates could be created or settings for public poll users)


dartcafe:
@v1r0x I tested uo and down. Will merge to get this into the master branch and prepare the RC-2 and the 1.6 release afterwards.

Issue #738, opened 2020-01-12
Limit numer of votes per participant


dartcafe:
Limit the numer of 'yes' votes a participant can set.

dartcafe:
DB is prepared

topiux:
@dartcafe +1 a must for our use case. Thanks for giving attention to this feature request.

christianlupus:
@dartcafe I would need this feature too. Is there any plan if and when this might get implemented in the frontend?
I am less concerned about a way to trick in multiple votes from one user but more of a UX such that there is only a set of radio buttons (aka exactly one option to choose from).

dartcafe:
No schedule right now. But not in the 1.5 version.

jhnet:
+1 as I would need this too ;)

annievoss:
+1 me too!

ostasevych:
Are any updates on this request?..

PR #1233, opened 2020-11-13
Adding new feature to polls

19Topgun93:
Would it be possible to add the feature that you can only select one choice? So that you have to choose one thing, and not the other options.

Best regrads


dartcafe:
@19Topgun93 We already have this feature request. See #738 Will be solved in a future release. 

Just for explanation, because this seems to be your first contribution: 
Pull requests are for merging branches not for bugs or requests. Please create an issue (see link above '[issues](https://github.com/nextcloud/polls/issues)') if you have a feature request or a bug report.

PR #1344, opened 2021-01-13
disallow commenting with public shares

dartcafe:
prevent commenting, when entering public poll without registration


tessus:
@dartcafe I don't understand this. It was very useful that people who did not have an account were able to comment on the poll. Now they have to call me or send a text or instant message. I'm very sorry, but this restriction makes no sense.

dartcafe:
Commenting is not removed. There was an error, that commenting was possible, without prior registration. This got fixed.

tessus:
Thank you very much for the explanation. From the title I could only see that now commenting is no longer possible with public shares, which would be a pitty. But apparenrly people can still comment when they do not have an account.

dartcafe:
Maybe this is a misleading title, but I didnt't expect someone to question this. 😄 
And the first post may be left unattended.

PR #1365, opened 2021-01-24
Date poll with time span and only day events

### LINKED ISSUE #192: Change in anonymous selection is not saved
### Steps to reproduce
1. save poll as anonymous
2. edit poll and remove anonymous
3. save it again

### Expected behaviour
Names should be visible now, or at least new voter names should be visible

### Actual behaviour
Poll is still anonymous, the change has not been saved

### Server configuration
**Nextcloud version:** (see Nextcloud admin page)
12

**Browser**
Firefox

If it is not reproducable I can add more informations

### LINKED ISSUE #201: Select date from to
Hi,

is it possible to add a feature to select date and time from ... to ...?
eg. if i will create a poll for a whole week I have to choose each day individually, but it would be nice if i click on mondey and slide to the friday

### LINKED ISSUE #404: Timespans and possibility to send appointments via mail
To organize internal trainings we want to start polls to give the employees the possibility to "book" the date they want.

At this moment there is no option to determine a time span. If there would be a chance to set a maximum number of participants (please take a look at my other feature request) it would be really great if the owner of the poll has the possibility to send invitations for the date to all participants in form of an outlook-appointment.

**Describe alternatives you've considered**
There are no alternatives accept the manual process.

### LINKED ISSUE #492: Add duration for events in the poll
When trying to create a poll for scheduling a meeing, often times the date and starting times are not enough if dealing with a number of invitees, whose schedule is very busy. These invitees often, or almost in our case, demand a precise duration for the event in question.

Provide a way to additionally set the duration for an event in the poll.

**Describe alternatives you've considered**
We tried text based polls, but these are not "calendarish" enough and people generally don't read all the text, if the poll is about finding a date for a meeting.

**Additional context**
Actually, doodle provides this feature and the adption will be increased and we can nudge peaople towards usingf the poll app on our NC instance, rather than doodle or the like.

### LINKED ISSUE #991: allow pure date polls without a time
### Is your feature request related to a problem? Please describe.
every now and then we need to find at least a date but don't know/need the exact time (yet).

### Describe the solution you'd like
an option to check, that no time is asked and shown.

**edit:** thinking a little bit further: perhaps this option also just per item would be interesting for some people...
- but that would make it necessary to add this option to the date/time-picker, otherwise it would be no pleasant UX (if you have to check the box for no time beforehand choosing the date).
- on the other hand adding a checkbox to "hide" the time for every chosen option would be a solution to change it afterwards... but this seems also a little bit clunky to me.

...so i'm not sure about a per item solution. :thinking: 

### Describe alternatives you've considered
- adding the time when it **probably** will take place, leads to a ton of discussions... :see_no_evil: 
  if it is clear, that the exact time is to be discussed/announced, some participants with *maybe* answers leave a comment when/why it is possible or not and that's it.
- also using 00:00 as time, raises sometimes confusions or even questions. :frowning:

### LINKED ISSUE #1118: 1.5RC: Horizontal scrolling fails with webkit browsers
With the RC of ver 1.5 scrolling horizontally moves the participants out of view. The participants start scrolling after more than 6 or 7 date options or more on wide screens / large browser windows.

Screenshot 2020-09-25 at 11 14 03

I experienced this issue with the current webkit browsers  Safari 14 on macOS and on iOS/iPadOS 14. Firefox and Chrome on Desktop (tested with Win10 and macOS) are OK.

#533 is back?

### LINKED ISSUE #1403: At choice resultfont and markoption in Darkmode isn't readable
### What is going wrong?
**Describe the bug**
Can't find the square to check the markoption below the table
If the result is elect by selectoption, the font and isn't readable. 

**To Reproduce**
1. Enable Darkmode in settings accessibility
2. Make a testpoll with yes and no choice
3. Elect yes
4. Close poll
5. Check mark the option yes below the table
6. yes isn't readable

**Expected behavior**
A dark font should show the table with yes 

**Screenshots**
Before:
image

After:
image

### Information about your polls installation
**Polls version?** (see apps page)
1.7.5

**How did you install this version?(Appstore or describe installation)**
By nextloud

### Information about your Instance of Nextcloud
Own server

**Nextcloud version:** (see Nextcloud admin page)
19.0.8

### Other informations

**Tested browser**
Firefox
Chromium

**Tested OS**
Debian
Mint

dartcafe:
Initial issues: 
* fix #492
* fix #201 
* fix #991 
* fix #404

* Options can now have a duration
* Options can be created only with a date or for multiple dates
* Realtime watching of polls enhanced
* fix #1403 - darkmode fix for selected options
* fix #1118 - should be fixed also (worked on iPadOs)

- [x] fix datepicker change from / to range mode
- [x] @nextcloud/designers Could need some design hints to improve usability (see updates in the next comment)


dartcafe:
### First implementation
grafik

1. Option without duration
2. Option without duration
3. Option with 1h duration (start and end same day)
4. Only day event with 24h duration
5. Only day event where the end day is another day (multiple day event) with 48h duration
6. Option with duration (start and end day are different) with 42h duration
7. Option with duration (start and end day are different) with 24h duration

### Date options can now have a duration
grafik
### Select time span via calendar component
grafik
grafik

Only (or multiple) day events are events starting at 0:00 local time with 24, 48, 72, ... duration. If the user's timezone differs from the owner's one, start and end time will get displayed. Same poll as above in a different time zone.
grafik

herrmannsdorfer:
I can confirm that #1118 is fixed with Polls 1.8.0b1 on macOS with Safari 14.0.3.

PR #1397, opened 2021-02-06
changed migrations for indices

dartcafe:
* handle indices in repair steps
* remove all indices in pre-migration repair step
* create all needed indices in post-migration repair step
* removed index creation from migration steps

Managing the creation of indices in the migration gets difficult and confusing. Therefore all indices get removed before a migration and get recreated after a successful migration. This way, we can handle all indices in one class and make sure, that the indices are set to the current database layout.


jotoeri:
Hey @dartcafe,
not sure if you recognized already, but i just wanted to install polls on my dev-setup and ran into an upgrade error. Seems to be connected to this PR!?
grafik

Greets 🙂 
Jonas

dartcafe:
Hmmm. Strange. All tests are running thru and I didn't have this issue. Normally this should work with the dependency injection. How did you install (update or new install) and which NC version? So thanks for reporting, but I have to inspect that.

jotoeri:
It was a NC 20.0.6, but persists after update to 20.0.7.
Initially i had some polls version from the app-store installed, then removed the folder and cloned the Repo. I ran `make dev-setup` and `npm run build`. The app-upgrade then shows the error.
Last migration, that ran was `0107Date20210121220707`.

PR #1405, opened 2021-02-12
l10n: Change quotation marks

Valdnet:


dartcafe:
The linter doesn't like them there, even escaped. That's why I used the single quotes.

Valdnet:
> That's why I used the single quotes.

Just as I thought 😉.

But you have to stick to certain standards in the application. If you use quotes `"` in other text strings in the application and in the Nextcloud instance, then they should be changed to valid. I don't know if the PR is correct. If something is in the way, correct me.

dartcafe:
@Valdnet I moved this to a property in #1441. I'll close this one. Thanks anyway.

PR #1469, opened 2021-03-07
change migration

dartcafe:
Changed migration for 1.8


dartcafe:
Made the migration valid for NC21, too.

PR #1490, opened 2021-03-17
Delete polls for deleted user

jotoeri:
Hey @dartcafe,
this is a first step to delete a users polls after the user got deleted.
I recently implemented this for forms and as i originally realised this for polls on the [forum](https://help.nextcloud.com/t/polls-update-to-new-version-fails/106410/3) and having nextcloud/forms#274 in mind, i thought why not just pushing the same here. 😉 
However:
- Currently this is not working, as the cron-job does not get the permission to delete on pollService
- I'm not sure, if this `$this->pollService->delete($poll->getId());` really deletes all information including options and votes for a poll. Probably somewhere is another procedure to delete all of that?

Unfortunately it's not just straight forward for me, so i'd need to dig deeper into the whole app. So if you want to use this part, just take it over (as far as i see, only the effective internal delete-action has to be fixed now). If you have too much other stuff around or don't want this, just close this PR. 😉 

Greets,
Jonas


dartcafe:
👍 That saved me some time 'cause that is still on my to do list. But so far I didn't dig into the event listner thing. Thanks. 
With the deletion of a poll, all child elements of that poll get deleted due to db constraints. 

But there is a little bit to do, we have to anonymize every action of the user, but votes should still stand after the user left the system.

jotoeri:
Yeah, i don't know which data you store additionally, but for forms i now left the submissions as is, with the thought that submissions, also in connection to a user-name, are 'property' of the form-owner once submitted. But that's indeed discussable and needs some handling of e.g. the DisplayName if a user can not be found anymore.

dartcafe:
I will complete this after 1.8 is out (estimated for Saturday).

dartcafe:
I did a rebase and switched the logic to delete via sql. 
@jotoeri Any reason, why you put this to a background job? Wouldn't it be better to delete the polls synchronously?

jotoeri:
Just that doing it synchronously slows down the user-deletion process... So if several apps insert their job synchronously that can become quite annoying. So afaik such things should always be done asynchronously. Iirc there also was some discussion on that on the community-chat somehwen...

dartcafe:
I see. I come over it, because the user gets deleted and his/her polls are still available with an invalid owner. This could confuse other users and looks like a bug.

jotoeri:
Well typically cron-jobs are run quite often (each five minutes typ. with cron?)... So i don't think that should be too much of a problem... 🤷‍♂️ 😉

dartcafe:
Finished the PR with deleting 
* shares
* preferences
* logs
* subscriptions
* shares

The userId in votes, comments and options will get renamed to a deleted status, to avoid changing poll options and results.

PR #1406, opened 2021-02-12
Fix test execution paths

kauron:
I could not get the tests to run correctly, but at least now the tests run. Fixes #1387 

In my case, they fail with `Error: Class 'OC' not found`, even when placed in the `apps` directory of an active Nextcloud installation. Could anyone point me to how to properly run the tests, or where the app should be placed to run these tests?

PR #1516, opened 2021-03-27
Add polls:share:* CLI commands

PhrozenByte:
Feedback is highly appreciated :heart:

* Invite users to polls using `polls:share:add`
* Remove user invitations from polls using `polls:share:remove`

Please note the following limitations for `polls:share:add`:

* Invitations emails are sent immediately, since we expect the poll to be created and set up first.
* No public shares. Couldn't think of a real-life use case unless we can create polls via CLI. Not sure whether this is even reasonable thinking about the required number of options...
* No display names for email invitations due to limitations in Symfony Console's API: We can't have multiple values per option (something like `--email jane@example.com "Jane Doe"` isn't possible) and a separate option will cause confusion (think of `--email jane@example.com --email john@example.com --email-name "John Doe"`: is `"John Doe"` the name for `jane@example.com`, or `john@example.com`? Symfony Console won't tell us). I was thinking about adding a syntax like `--email "John Doe <john@example.com>"`, however, I don't think that this syntax should be exclusive to the CLI. What do you think about adding this to the Web UI?
* No invitations by circle via CLI. This isn't worth the effort, because circles are basically unusable via CLI due to API limitations anyway. See some reasoning [here](https://github.com/nextcloud/spreed/pull/3505#issuecomment-631969201) and [here](https://github.com/nextcloud/spreed/pull/3505#issuecomment-632061351).
* No invitations by contacts and contact groups, because contacts are basically content of the current user and there's no current user in CLI commands. Similar reason like with circles. Use `--email` instead.

Also note the following limitation for `polls:share:remove`:

* One can't remove public shares, because we can't identify public shares via CLI. However, the shares created by participants (participants create their own share when visiting a public share) can be removed using the `--email` option.
* Again, no support for circles. See above.
* One can't remove contact group invitations - mostly for the same reason as for `polls:share:add`. Contact invitations are treated just like email invitations and can be removed using the `--email` option (if the email address is known).


dartcafe:
@PhrozenByte I will come back to this later. Sorry for delay... 
But I am curious, what the use case is for this.

PhrozenByte:
Inviting a lot of people via e-mail :smile:

dartcafe:
> 
> 
> Inviting a lot of people via e-mail 😄

This is already possible. 😁 My question addressed the use case or better; a user story.

* Why or in which situation should one invite people via occ command? 
* Is the use case common or a special one? 
* Has the CLI way a valuable  advantage over using the GUI or the REST-Interface?
* Who would use that? 

Can we tell it as a user story?

PhrozenByte:
"I've a huge (usually line separated) list of e-mail addresses and want to invite these people, possibly at a later, time-controlled moment (i.e. via cronjob)". The GUI is way to clumsy to invite a few hundred users when they aren't already somehow stored in Nextcloud.

dartcafe:
Hmmm. I see. But I am not sure, if this would be a widly used feature. I see a problem in getting this maintained in the future, as we will not test this when a new release will get out. 

So we would need some test automation here at least.

PhrozenByte:
Agreed. I'll add some integration tests in the next couple of days.

dartcafe:
Cool. I merged #1406, which has changes in the test setup. Maybe a rebase is a good idea.

PhrozenByte:
Timing :laughing: I was wondering yesterday, too. Will do a rebase :+1:

tests/Integration/Command/Share/TShareCommandTest.php
```diff
@@ -0,0 +1,129 @@
...
+			->disallowMockingUnknownTypes()
+			->disableAutoReturnValueGeneration();
+
+		if ($addMethods !== null) {
+			$mockBuilder->addMethods($addMethods);
+		}
+
+		if ($onlyMethods !== null) {
+			$mockBuilder->onlyMethods($onlyMethods);
+		}
```
PhrozenByte, line 125: By the way: I'd highly recommend not to use magic methods for getters/setters. Explicitly adding the methods to mock is only necessary because of magic methods.

PhrozenByte:
Here we go :+1:

PhrozenByte:
Any updates @dartcafe?

dartcafe:
Sorry, not yet. I had to finish some open construction sites, but have a rather clean desk now. 

But one question: I avoided using the notificationService inside the shareServices to avoid circled dependencies. In your solution the shareService now depends on the notificationService and the notificationService depends on the shareService. Isn't this a problem?

tests/Integration/Command/Share/TShareCommandTest.php
dartcafe: I have no experience with this. The code for the tests isn't created by me. Do you mind to optimize that?

PhrozenByte:
A circular dependency would be a problem, indeed. However, `NotificationService` isn't depending on `ShareService`? Neither directly, nor via its only dependency, `NotificationManager`. Or do I miss something?

tests/Integration/Command/Share/TShareCommandTest.php
PhrozenByte: This is not about the test's code :smile:

The issue resides within your model classes. Just take the `OCA\Polls\Db\Share::getId()` method as an example: There's no `OCA\Polls\Db\Share::getId()` method, there's only a `OCP\AppFramework\Db\Entity::__call()` method that "magically" returns the requested property. Using magic methods is discouraged, because they are incredibly hard to test. This code snippet (also see [[1]](https://github.com/nextcloud/polls/pull/1516/files#diff-6a85ee1ca2ddc000e18ce3f0c0a8d2104197e04d4e0badb9a5c9effbff294462R66) and [[2]](https://github.com/nextcloud/polls/pull/1516/files#diff-6a85ee1ca2ddc000e18ce3f0c0a8d2104197e04d4e0badb9a5c9effbff294462R79-R80)) works around this issue by manually specifying all methods to mock - not just the magic methods, but *all* methods. This is bad. It works, but is nevertheless bad. It's the only solution because you're using magic methods. You should rather implement the getters you need (i.e. e.g. `OCA\Polls\Db\Share::getId()`).

Anyway, this is just a heads up for your. Fixing this technical debt is not in the scope of this PR. I can open a separate issue if you want to.

dartcafe: Ah. I thought it was about this implementation: 
https://github.com/nextcloud/polls/blob/09b4c9006b070ae7e36f9f4cf947a02069980d45/tests/Unit/FactoryMuffin.php#L19-L32

The style of magic getters and setters follows the NC style guide IMHO. Wasn't questioned by me. 

https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/database.html#entities

THB, this project was my first bigger developing experience.

PhrozenByte: Another workaround for magic getters/setters :smile: This is nothing that needs to be addressed *right now*, you just might want to keep it in mind.

dartcafe:
Idon't know, what I saw, but it was nonsens. 😕

dartcafe (dismissed):
Seems to be OK. Need some free time to review that PR. I didn't review the tests by detail. Just make the linter happy, and we can merge...

PhrozenByte:
> Just make the linter happy, and we can merge...

I'd say that this is a false positive. Anyway, done :+1:

dartcafe:
> 
> 
> > Just make the linter happy, and we can merge...
> 
> I'd say that this is a false positive. Anyway, done 👍

Yes it is. Anyway, done also....

dartcafe:
And thanks for your contribution

PR #1569, opened 2021-04-26
l10n: Correct and add a word

Valdnet:

### DIFF
--- a/src/js/components/Poll/PublicRegisterModal.vue
+++ b/src/js/components/Poll/PublicRegisterModal.vue
@@ -144,12 +144,12 @@ export default {
 				}
 			} else if (!this.isValidName) {
 				return {
-					result: t('polls', 'Invalid name'),
+					result: t('polls', 'Invalid username'),
 					status: 'error',
 				}
 			}
 			return {
-				result: t('polls', '{username} is valid.', { username: this.userName }),
+				result: t('polls', 'Username {username} is valid.', { username: this.userName }),
 				status: 'success',
 			}
 
--- a/src/js/components/Poll/PublicRegisterModal.vue
+++ b/src/js/components/Poll/PublicRegisterModal.vue
@@ -128,28 +128,28 @@ export default {
 		userNameCheck() {
 			if (this.checkingUserName) {
 				return {
-					result: t('polls', 'Checking username …'),
+					result: t('polls', 'Checking name …'),
 					status: 'checking',
 				}
 			}
 			if (this.userName.length === 0) {
 				return {
-					result: t('polls', 'Enter a username to participate.'),
+					result: t('polls', 'Enter a name to participate.'),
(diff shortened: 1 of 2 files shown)


src/js/components/Poll/PublicRegisterModal.vue
```diff
@@ -139,17 +139,17 @@ export default {
...
 			} else if (this.userName.length < 3) {
 				return {
-					result: t('polls', 'Please use at least 3 characters.'),
+					result: t('polls', 'Username must be at least 3 characters.'),
 					status: 'error',
 				}
 			} else if (!this.isValidName) {
 				return {
-					result: t('polls', 'Invalid name'),
+					result: t('polls', 'Invalid username'),
```
dartcafe, line 147: We avoided the usage of the term username, because some users were confused and entered their username, which they use to login and this lead to a validation error.

dartcafe (changes requested):

src/js/components/Poll/PublicRegisterModal.vue
Valdnet: This in such a case should also be changed in other text strings, such as:
https://github.com/nextcloud/polls/blob/ad6e46e17406f12e4a8f6c1443841dbd0108d7b5/src/js/components/Poll/PublicRegisterModal.vue#L142

Valdnet: This in such a case should also be changed in other text strings, such as:
https://github.com/nextcloud/polls/blob/ad6e46e17406f12e4a8f6c1443841dbd0108d7b5/src/js/components/Poll/PublicRegisterModal.vue#L137
e.t.c.

dartcafe: Absolutely. Would you mind correcting this, too?

Valdnet: Changed 😉. Can you approve?

dartcafe (approved):

src/js/components/Poll/PublicRegisterModal.vue
```diff
@@ -149,7 +149,7 @@ export default {
 				}
 			}
 			return {
-				result: t('polls', '{username} is valid.', { username: this.userName }),
+				result: t('polls', 'Name {username} is valid.', { username: this.userName }),
```
dartcafe, line 152: Sounds strange (Name Peter is valid). Term should be kept short.

dartcafe (changes requested):

dartcafe:
Ah. You merged already... 👍

src/js/components/Poll/PublicRegisterModal.vue
Valdnet: You're right 😉. Changed in #1573.

PR #2057, opened 2021-10-29
Changed grammar

rakekniven:
Changed grammar to align to other strings in NC project.

Reported at Transifex.

Signed-off-by: rakekniven <2069590+rakekniven@users.noreply.github.com>


dartcafe:
I added the introduction from past tense to present tense, so the present perfect make more sense.

I additionally changed some translations strings, which have been introduced recently and removed the prefixes, to avoid new translations, if they get removed later.

dartcafe (approved):

dartcafe:
I hope, this isn't too confusing...

PR #2155, opened 2021-12-29
l10n: Correct spelling

Valdnet:
Reported by the user from Transifex.

### DIFF
--- a/lib/Service/ActivityService.php
+++ b/lib/Service/ActivityService.php
@@ -94,12 +94,12 @@ class ActivityService {
 		switch ($this->eventType) {
 			case CommentEvent::ADD:
 				return $this->userIsActor
-					? $this->trans->t('You commented poll {pollTitle}')
-					: $this->trans->t('{actor} commented poll {pollTitle}');
+					? $this->trans->t('You commented on poll {pollTitle}')
+					: $this->trans->t('{actor} commented on poll {pollTitle}');
 			case CommentEvent::DELETE:
 				return $this->userIsActor
-					? $this->trans->t('You deleted a commented from poll {pollTitle}')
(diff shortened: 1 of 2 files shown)


dartcafe:
@Valdnet Great, thanks. 👍  According to #2057 we should use the present perfect here.

lib/Service/ActivityService.php
```diff
@@ -355,7 +355,7 @@ private function getMessageFiltered() {
 
 				return $this->userIsActor
 					? $this->trans->t('You deleted {sharee}\'s share')
-					: $this->trans->t('{actor} deleted a share');
```
dartcafe, line 358: Avoiding the sharee here was by intention, for possible privacy concerns.

dartcafe (changes requested):

lib/Service/ActivityService.php
Valdnet: I have restored the text string 😉.

dartcafe (approved):

lib/Service/ActivityService.php
```diff
@@ -187,7 +187,7 @@ private function getMessageFull() {
 			case ShareEvent::REGISTRATION:
 				return $this->userIsActor
 					? $this->trans->t('You registered to poll {pollTitle}')
-					: $this->trans->t('{sharee} registered to poll {pollTitle}');
+					: $this->trans->t('{actor} registered to poll {pollTitle}');
```
dartcafe, line 190: In this case {sharee} was right to fetch the correct user type, when displaying the user's name

dartcafe (changes requested):

dartcafe:
@Valdnet 😆 Too fast. The review approvement was related to the change. I will change the wordings in the next hour and fix the last change request in a another PR. Thanskf or review and correcting.

PR #2171, opened 2022-01-07
l10n: Fixed grammar and changed wording of genitive case

rakekniven:
Reported at Transifex.

Signed-off-by: rakekniven <2069590+rakekniven@users.noreply.github.com>

### DIFF
--- a/lib/Service/ActivityService.php
+++ b/lib/Service/ActivityService.php
@@ -233,7 +233,7 @@ class ActivityService {
 
 		return $this->userIsActor
 			? $this->trans->t('You have done something indescribable with poll {pollTitle}')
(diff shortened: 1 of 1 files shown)


dartcafe:
@rakekniven What is the reason for using of the `of genitive` over the `'s genitive`? 
Problems with names ending with s?

rakekniven:
> @rakekniven What is the reason for using of the `of genitive` over the `'s genitive`? Problems with names ending with s?

In the past there was some discussion at Transifex. Cannot remember the string to provide a link ... 🤔 

For me this also looks better in the activity list because the variable is mostly displayed in bold. The "s" would be shown normal.

Issue #1986, opened 2021-09-28
Mixed Poll for different subgroups of a large group


Elvis2000NC:
I would like to explain the idea with my example. I play in a big band and we have different kinds of rehearsals. Some days only the trumpeters rehearse, some days the trombones rehearse and then sometimes all the musicians rehearse. Now, as the band leader, I want to have an overall view of who is rehearsing and when. So I would like to put all the dates in only one survey. NOW, however, I would like to assign the dates to different user groups (trumpet, trombone, all). When a trumpet player logs in via his account, he should only see the dates that are relevant to him. If a trombonist logs in, he should only see his appointments for the survey. I, as the band leader, see all the dates (perhaps highlighted in a different colour), but I can also show and hide the individual dates of the groups.

dartcafe:
@Elvis2000NC  I had a similair use case. See the linked PR, which added a view, where multiple date polls can be viewed in one view. Maybe this solves your use case too.

Elvis2000NC:
that looks fine, but how do you make it? i try it in my polls, but i don't find the option "select polls to combine"?
i have two different polls.
or do you have one poll that you separated first and combine again?

dartcafe:
This feature is planned for the next version and still in development.

Elvis2000NC:
ah, perfect. Thankx
   
  
  
   
  
  
   
    René Gieling ***@***.***> hat am 10.01.2022 08:42 geschrieben:
   
   
    
   
   
    
   
   
   This feature is planned for the next version and still in development.
   —Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android. You are receiving this because you were mentioned.Message ID: ***@***.***>
  
  
   
  
  
   ____________________________________________________________
   DOCTORES KAUL
   Dr. Alexander Kaul
   Dr. Katharina Kaul, M.Sc. Kinderzahnheilkunde
   Schleidener Str. 120
   52076 Aachen
   Tel. 02408 80204
   Fax 02408 80225
   Email: ***@***.***
   www.doctores-kaul.de

PR #2175, opened 2022-01-09
adding a combined day polls view

dartcafe:
Added a function, to be able to get a combined view for multiple polls. 

## Example adopted from #1986, given two polls:
first poll | second poll
-|-
grafik |  grafik

## Select the polls to combine in the sidebar
grafik

## resulting in 
grafik


Valdnet (dismissed):

Valdnet:
I think it's great 👍.

Valdnet (approved):

PR #2294, opened 2022-02-26
Move title to header and rearrange action buttons

dartcafe:
* Added a header area for the poll title
* moved sorting and view switches close to the vote table 
* add id modifier to markdown area to avoid conflicts with core styles

before | after |
-- | -- | 
grafik | grafik | 
grafik | grafik |
grafik | grafik |


dartcafe:
@nextcloud/designers I could need some feedback or suggestions 

Goals: 
* Make clearer that the sorting and view buttons affect the vote table only
* Move the title to the header together with the remaining buttons
* leave space for the navigation toggle

PR #2351, opened 2022-04-03
store personal token for a public poll in a cookie

dartcafe:
For a better UX a public user can decide, if he wants to keep the private token for a public poll in a cookie.
The cookie expiry is currently set to 30 days and will get renewed after every access.

grafik

- [x] Consider to renew cookie expiry after every access and reduce expiry duration.

### DIFF
--- a/src/js/components/Poll/PublicRegisterModal.vue
+++ b/src/js/components/Poll/PublicRegisterModal.vue
@@ -35,6 +35,9 @@
 							no-submit
 							focus
 							@submit="submitRegistration" />
+						<CheckboxRadioSwitch :checked.sync="saveCookie">
+							{{ t('polls', 'Save username in cookie for 30 days.') }}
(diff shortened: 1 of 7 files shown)


Valdnet (approved):

src/js/components/Poll/PublicRegisterModal.vue
```diff
@@ -35,6 +35,9 @@
 							no-submit
 							focus
 							@submit="submitRegistration" />
+						<CheckboxRadioSwitch :checked.sync="saveCookie">
+							{{ t('polls', 'Save username in cookie for 30 days.') }}
```
Valdnet, line 39:
```suggestion
							{{ t('polls', 'Save username in cookie for 30 days') }}
```

Valdnet (approved):

Issue #2430, opened 2022-05-22
Wrong translation in polish language


Maniok2212:
## What is going wrong? What did you observe?
**Wrong translation in "enter name" dialogue**

It literary says "name is not correct" in polish

![IMG_20220522_125534](https://user-images.githubusercontent.com/106015493/169700134-0a997908-f3a1-449a-a1d7-dbbd8b416311.jpg)

## How should it work?
"nieprawidłowa" means "not correct". It should be "prawidłowa" which means "correct".

## Information about your polls installation
**Polls version:** 3.6.1

**Fresh installation**

- [x] First time install
- [ ] Update from a prior version
  Version number of the previous version:

**How did you install this version?(Appstore or describe installation)**

- [x] Installed from the appstore
- [ ] Installed via occ
- [ ] Installed via extracting downloaded package to the apps folder

## Information about your Instance of Nextcloud
**Nextcloud version:** 24.0.1

</details>

dartcafe:
@Valdnet Can you check that in transifex?

Valdnet:
@Maniok2212 Thanks for reporting a translation error 👍. The text string was incorrectly translated a year ago by a translator from Transifex.
I also invite you to check and actively translate other new text strings 😉.

@dartcafe The text string has been changed to look more stylistic.
Please approve #2443.

PR #2443, opened 2022-05-29
l10n: Add word "Name"

Valdnet:
Fix issue #2430.

### DIFF
--- a/src/js/components/Poll/PublicRegisterModal.vue
+++ b/src/js/components/Poll/PublicRegisterModal.vue
@@ -190,13 +190,13 @@ export default {
 
 			if (!this.isValidName) {
 				return {
-					result: t('polls', '{username} is invalid or reserved.', { username: this.userName }),
+					result: t('polls', 'Name {username} is invalid or reserved.', { username: this.userName }),
 					status: 'error',
 				}
 			}
 
 			return {
-				result: t('polls', '{username} is valid.', { username: this.userName }),
+				result: t('polls', 'Name {username} is valid.', { username: this.userName }),
 				status: 'success',
 			}
 		},
--- a/src/js/components/Poll/PublicRegisterModal.vue
+++ b/src/js/components/Poll/PublicRegisterModal.vue
@@ -190,13 +190,13 @@ export default {
 
 			if (!this.isValidName) {
 				return {
-					result: t('polls', 'User name {username} is invalid or reserved.', { username: this.userName }),
+					result: t('polls', 'The name {username} is invalid or reserved.', { username: this.userName }),
 					status: 'error',
 				}
 			}
 
 			return {
-				result: t('polls', 'User name {username} is valid.', { username: this.userName }),
+				result: t('polls', 'The name {username} is valid.', { username: this.userName }),
 				status: 'success',
 			}
 		},
--- a/src/js/components/Poll/PublicRegisterModal.vue
+++ b/src/js/components/Poll/PublicRegisterModal.vue
@@ -190,13 +190,13 @@ export default {
 
 			if (!this.isValidName) {
 				return {
-					result: t('polls', 'Name {username} is invalid or reserved.', { username: this.userName }),
+					result: t('polls', 'User name {username} is invalid or reserved.', { username: this.userName }),
 					status: 'error',
 				}
 			}
 
 			return {
-				result: t('polls', 'Name {username} is valid.', { username: this.userName }),
+				result: t('polls', 'User name {username} is valid.', { username: this.userName }),
 				status: 'success',
 			}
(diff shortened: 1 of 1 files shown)


src/js/components/Poll/PublicRegisterModal.vue
```diff
@@ -190,13 +190,13 @@ export default {
 
 			if (!this.isValidName) {
 				return {
-					result: t('polls', '{username} is invalid or reserved.', { username: this.userName }),
+					result: t('polls', 'Name {username} is invalid or reserved.', { username: this.userName }),
```
dartcafe, line 193: "Name Peter is invalid" sounds strange to me. Either it needs an article ("The name Peter is invalid") or the term stays as before ("Peter is invalid").

src/js/components/Poll/PublicRegisterModal.vue
```diff
@@ -190,13 +190,13 @@ export default {
...
 					status: 'error',
 				}
 			}
 
 			return {
-				result: t('polls', '{username} is valid.', { username: this.userName }),
+				result: t('polls', 'Name {username} is valid.', { username: this.userName }),
```
dartcafe, line 199: Dito. In doubt we could remove this and the 'Checking name ...' as well, return an empty string as default and skip the information. This reduces the noise.

dartcafe (changes requested):

src/js/components/Poll/PublicRegisterModal.vue
rakekniven: "User name **Peter** is invalid" would be my suggestion.

src/js/components/Poll/PublicRegisterModal.vue
rakekniven: Where does this message appears? As a Pop up? Or below a form item?

Valdnet: > In doubt we could remove this and the 'Checking name ...' as well, return an empty string as default and skip the information. This reduces the noise.

@dartcafe What about a wrong username?

> Where does this message appears?

@rakekniven See issue # 2430.

rakekniven (approved):

src/js/components/Poll/PublicRegisterModal.vue
dartcafe: @rakekniven We had feedback, that the term `username` could be misleading for users who visit a public poll and are also a registered user of this site. This was the initial reason, why the term `username` was removed.

src/js/components/Poll/PublicRegisterModal.vue
```diff
@@ -190,13 +190,13 @@ export default {
 
 			if (!this.isValidName) {
 				return {
-					result: t('polls', 'Name {username} is invalid or reserved.', { username: this.userName }),
+					result: t('polls', 'User name {username} is invalid or reserved.', { username: this.userName }),
```
dartcafe, line 193: Please see my last comment. User name or username should be avoided in this combination.

dartcafe (changes requested):

src/js/components/Poll/PublicRegisterModal.vue
dartcafe: > What about a wrong username? 

@Valdnet See also my last [comment](https://github.com/nextcloud/polls/pull/2443#discussion_r884582223).

src/js/components/Poll/PublicRegisterModal.vue
rakekniven: Ok, good catch.

So what about "Participant name **Peter** is invalid or reserved"?

src/js/components/Poll/PublicRegisterModal.vue
rakekniven: See my comment above.

src/js/components/Poll/PublicRegisterModal.vue
Valdnet: @dartcafe So it should be: `User name is invalid or reserved.`?

dartcafe: @Valdnet 

Username can be interpreted as a user id, which is the login name of a user. In this case we just ask for a random name, therefore is should only be called `name`, IMHO.

It is just "Name Peter is invalid or reserved." still sounds strange to me, why I prefer to say "The name Peter is invalid or reserved."

dartcafe: That's what I think fits best, if the old variant isn't clear enough fot the translations.

rakekniven (approved):

PR #2559, opened 2022-08-26
Fix darkmode in NC25+

dartcafe:
* change css due to new css variables since NC25
* handle prefer-color-scheme in public polls and new apperance setting

ref https://github.com/nextcloud/server/issues/32060

### DIFF
--- a/src/js/assets/scss/colors-dark.scss
+++ b/src/js/assets/scss/colors-dark.scss
@@ -1,8 +1,11 @@
 // TODO: preparation for full support of @media:prefers-color-scheme by core
 // until then, this ha no effect
 
+// dark mode, when
+// @media (prefers-color-scheme: dark) && [data-themes="default"]
+
 @media (prefers-color-scheme: dark) {
-	:root {
+	:root, [data-theme-default] {
 		--color-foreground-error: #dddddd;
 		--color-foreground-success: #dddddd;
 		--color-background-success: #2e4600;
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -23,7 +23,7 @@
     <screenshot>https://raw.githubusercontent.com/nextcloud/polls/master/screenshots/edit-poll.png</screenshot>
     <dependencies>
         <php min-version="7.4"/>
-        <nextcloud min-version="22" max-version="24" />
+        <nextcloud min-version="25" max-version="25" />
     </dependencies>
     <activity>
 		<providers>
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -23,7 +23,7 @@
     <screenshot>https://raw.githubusercontent.com/nextcloud/polls/master/screenshots/edit-poll.png</screenshot>
     <dependencies>
         <php min-version="7.4"/>
-        <nextcloud min-version="25" max-version="25" />
+        <nextcloud min-version="22" max-version="25" />
     </dependencies>
     <activity>
 		<providers>
--- a/src/js/assets/scss/colors.scss
+++ b/src/js/assets/scss/colors.scss
@@ -1,3 +1,4 @@
+// TODO: Remove :root selector after NC24
 :root {
 	--color-foreground-error: #222222;
 	--color-foreground-success: #222222;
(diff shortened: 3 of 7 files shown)


dartcafe:
Thereare some ugly tweeks, but it works for now.

src/js/assets/scss/colors-dark.scss
```diff
@@ -1,8 +1,11 @@
 // TODO: preparation for full support of @media:prefers-color-scheme by core
 // until then, this ha no effect
 
+// dark mode, when
+// @media (prefers-color-scheme: dark) && [data-themes="default"]
+
 @media (prefers-color-scheme: dark) {
-	:root {
+	:root, [data-theme-default] {
```
skjnldsv, line 8: If new themes gets added, your app will break

dartcafe: @skjnldsv Hi, thanks, but as I said in the comments: "Some dirty tweeks." 😉 

The quandary is, to detect if the user opened a public page in NC<25 or an internal page in NC25 with system detect theme. Any suggestion?

skjnldsv: Not sure I get it, what is this about? :)

skjnldsv: If you need to know if the page is public or not, do it via PhP and check the rendered template?

dartcafe: like a miracle I found `body-public`. But the proble are not the public pages. I have to tell if internal pages ar from NC22 or NC25 with system theme detection. The only indicator to me seems to be `[data-theme-default]` to detect, that this page is from NC25 with activated system theme detection.

skjnldsv: Why do you make a difference between 22 and 25 ?
The variables did not changed. We just extended the dark mode. :thinking:

dartcafe: Nope. NC25 introduced the system theming setting also for internal pages. So I cannot use `prefers-color-scheme: dark` without making sure that the user has theming settings to `System default setting` (which seems to be represented by `[data-theme-default]`). 

If the user has the light theme configured in NC but a dark system theme, the page would be rendered with light variables, but the app individual variables would apply the setting from `prefers-color-scheme`. 

For public pages, this is easy, because public users have no theming setting and the system setting gets applied since a coule of versions.

nursoda: @dartcafe You distinguish two cases:

1. a user is logged in – thus NC preferences shall be honored (overriding possibly different browser/system theme)
2. no NC user session – thus NC preferences cannot and don't have to be honored, always use browser/system theme

This applies to public/link shared files pages just the same as to public polls, doesn't it?
So how does the core file sharing functionality do it?

I'm still struggling to find out what this all is about and where to start. Since I'm not a dev, maintaining twofactor_email just got even more challenging. If you're interested in mutual exchange (probably more helping me out), feel free to contact me. see privat seyfarth de olav.

dartcafe: The public pages are not the problem, because the system setting is respected since a couple of versions. 
The problem are the internal pages, because we have some colors defined, which have to be set according to the theme.

And we (better I) want to keep the compatibility to NC prior to 25. 

So when it comes to NC24 and before, we have to select the colors according to the default (light) theme or the dark theme. easy

Since 25, we have a third situation. The (logged in) user can have set a dark or light theme, which means the system setting is to be ignored. But if he has set the system theme, `prefers-color-scheme: dark` comes into play. But is has only to be applied, if the user has set this theme, not whan dark or light theme is selected. 

I found out that the system theme seems to be indicated by the body attribute `default-theme`. Maybe a misinterpartion, but it seems to work as expected. Therefore I chosed these combinations. 

|  | logged in user dark theme set | logged in user light theme set | logged in user system theme set |
|-|-|-|-|
|NC 25 | ignore `prefers-color-scheme:dark`, use `[light-theme]` | ignore `prefers-color-scheme:dark`, use `[light-theme]` | apply prefers-color-scheme, use `[default-theme]`|
NC 24 | ignore `prefers-color-scheme`, use `body.theme-dark`| ignore `prefers-color-scheme`, use `body.theme-light`| not valid |

**Fallback then are light colors.**

So it comes that `@media:prefers-color-scheme: dark` only has to be applied, if
* user is not logged in OR
* user is logged in 
  * AND NC is >24  
  * AND user has not set the light theme --> (current indicator `[default-theme]` exists)

dartcafe: To tell if a page is apublic page or not can be told by the body id
* internal pages have the id `body#body-user`
* public pages have the id `body#body-public`

I don't know since when, but i discovered this while fiddling out the theme differences.

dartcafe: yes. copy&paste error. Thanks. 🙈

PR #2734, opened 2023-01-10
fix/refactor acl

dartcafe:
fixing user rights and refactor access validation checks

### DIFF
--- a/lib/Model/Acl.php
+++ b/lib/Model/Acl.php
@@ -163,11 +163,11 @@ class Acl implements JsonSerializable {
 
 	public function getPollId(): int {
 		return $this->poll->getId();
-	}	
+	}
 
 	public function getToken(): string {
 		return strval($this->share->getToken());
-	}		
+	}
 
 	public function getUserId(): string {
 		return $this->getIsLoggedIn() ? $this->userSession->getUser()->getUID() : $this->share->getUserId();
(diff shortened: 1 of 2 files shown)


lib/Model/Acl.php
```diff
@@ -489,4 +443,170 @@ private function getIsShareValidForUsers(): bool {
 	private function getHasEmail(): bool {
 		return $this->share->getToken() ? strlen($this->share->getEmailAddress()) > 0 : $this->getIsLoggedIn();
 	}
+
+	/**
+	 * Checks, if user is allowed to edit the poll configuration
+	 **/
+	private function getAllowEditPoll(): bool {
+		// Console god mode
+		if (defined('OC_CONSOLE')) {
```
nickvergessen, line 452: There is also `OC::$CLI`
But that is also true for system cron jobs.

lib/Model/Acl.php
```diff
@@ -489,4 +443,170 @@ private function getIsShareValidForUsers(): bool {
...
+		if (!$this->getAllowAccessPoll()) {
+			return false;
+		}
+
+		// public shares are not allowed to comment
+		if ($this->share->getType() === Share::TYPE_PUBLIC) {
+			return false;
+		}
+
+		return !!$this->poll->getAllowComment();
```
nickvergessen, line 564: That is very "JS" like. In PHP you would normally do:
```suggestion
		return (bool) $this->poll->getAllowComment();
```

but both works.

nickvergessen (approved):

lib/Model/Acl.php
dartcafe: Ah. Thanks!

lib/Model/Acl.php
dartcafe: I try to avoid OC if possible, since it is not defined on dev time.

nickvergessen (approved):

PR #2840, opened 2023-04-03
Typo swinwn

ThomasKujawa:
from social.tchncs.de/@nahtanoy


dartcafe:
Thanks, but this has to get fixed in [Transifex](https://app.transifex.com/nextcloud/nextcloud/translate/#de/polls/405711017). I fixed that.

PR #2843, opened 2023-04-03
Do not try to query unsupported interface class via app-container

### LINKED ISSUE #2842: 5.0.0-rc1: IEMailTemplate class cannot be queried via container
The following just cannot work, as there is no registered service for the IEMailTemplate class in the app container:

https://github.com/nextcloud/polls/blob/b785f16c1fc6eeaa35a755b9ebf2ceaa46914142/lib/Model/Mail/MailBase.php#L74

Or is it my setup? I constantly get  `OCP\\AppFramework\\QueryException`  when this interface class is queried.

This happens with the forthcoming v5 release.

rotdrop:
Potentially fixes #2842


dartcafe:
Thanks, but already fixed in  #2836.

PR #2950, opened 2023-06-20
fix innerJoin with PostgeSQL

dartcafe:
fixes #2946
* fix innerJoin in TableManager
* add rebuild command to server setup action

PR #2952, opened 2023-06-21
fix(deleteDuplicates): select only duplicates

hrenard:
Hey,

@unteem caught a typo you made in https://github.com/nextcloud/polls/pull/2950 which cause all rows de be deleted.

### DIFF
--- a/lib/Db/TableManager.php
+++ b/lib/Db/TableManager.php
@@ -319,7 +319,7 @@ class TableManager {
 			// identify duplicates
 			$selection = $qb->selectDistinct('t1.id')
 			->from($table, 't1')
-			->innerJoin('t1', $table, 't2', $qb->expr()->eq('t1.id', 't2.id'));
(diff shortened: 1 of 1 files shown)


lib/Db/TableManager.php
```diff
@@ -319,7 +319,7 @@ private function deleteDuplicates(string $table, array $columns):int {
 			// identify duplicates
 			$selection = $qb->selectDistinct('t1.id')
 			->from($table, 't1')
-			->innerJoin('t1', $table, 't2', $qb->expr()->eq('t1.id', 't2.id'));
+			->innerJoin('t1', $table, 't2', $qb->expr()->lq('t1.id', 't2.id'));
```
dartcafe, line 322: Should be ->lt() 😄

dartcafe (changes requested):

lib/Db/TableManager.php
hrenard: Oups sorry :upside_down_face:

dartcafe (approved):
Great. Thanks!

PR #2985, opened 2023-07-09
use $schemaClosure in migration

### LINKED ISSUE #2980: App doesn't cooperate with `occ db:convert-type`

- [ ] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

When converting to another db type, the tables are not created and therefore not converted.

In

https://github.com/nextcloud/polls/blob/0a8c7451b33c962db01ad65599ca5166421096ce/lib/Migration/Version050100Date20230515083001.php#L48

`$schemaClosure` is not used and therefore migration to a different schema/database is currently not supported.

Well, the app should cooperate by using `$schemaClosure` and return the schema (OC\DB\SchemaWrapper) the closure returns. See https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/migrations.html
If it does that, it shouldn't do
https://github.com/nextcloud/polls/blob/0a8c7451b33c962db01ad65599ca5166421096ce/lib/Migration/Version050100Date20230515083001.php#L51
since Nextcloud`s MigrationService takes care of the rest (e.g. uses the correct connection). I have tested this and it fixes the issue.

1. use `occ db:convert-type`
2. observe rows `oc_polls_...` under "The following tables will not be converted:"

### Installation method

Installed/updated from the appstore (Apps section of your site)

### Installation type

First time installation

### Affected polls version

5.1.0

- [X] Firefox
- [ ] Chrome
- [ ] Chromium/Chromium based (i.e. Edge)
- [ ] Safari
- [ ] Other/Don't know

### Other browser

_No response_

### Add your browser log here

_No response_

_No response_

### NC version

Nextcloud 26 (Nextcloud Hub 4)

### Other Nextcloud version

_No response_

### PHP engine version

PHP 8.1

### Other PHP version

_No response_

### Database engine

MariaDB

11.0.2

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other/Don't know

_No response_

_No response_

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Additional Information

_No response_

notEvil:
fixes #2980

Please note that I'm not particularly familiar with php and the codebase in general. The solution is a proposal.

### DIFF
--- a/lib/Db/TableManager.php
+++ b/lib/Db/TableManager.php
@@ -156,18 +156,22 @@ class TableManager {
 	 *
 	 * @psalm-return non-empty-list<string>
 	 */
-	public function createTables(): array {
+	public function createTables(?Schema $schema = null): array {
+		if ($schema === null) {
+			$schema = &$this->schema;
+		}
+
 		$messages = [];
 		
 		foreach (TableSchema::TABLES as $tableName => $columns) {
 			$tableName = $this->dbPrefix . $tableName;
 
-			if ($this->schema->hasTable($tableName)) {
-				$table = $this->schema->getTable($tableName);
+			if ($schema->hasTable($tableName)) {
+				$table = $schema->getTable($tableName);
 				$messages[] = 'Validating table ' . $table->getName();
 				$tableCreated = false;
 			} else {
-				$table = $this->schema->createTable($tableName);
+				$table = $schema->createTable($tableName);
 				$tableCreated = true;
 				$messages[] = 'Creating table ' . $table->getName();
 			}
--- a/lib/Migration/Version050100Date20230515083001.php
+++ b/lib/Migration/Version050100Date20230515083001.php
@@ -47,14 +47,14 @@ class Version050100Date20230515083001 extends SimpleMigrationStep {
 	 */
 	public function changeSchema(IOutput $output, \Closure $schemaClosure, array $options) {
 		// Create tables, as defined in TableSchema or fix column definitions
-		$messages = $this->tableManager->createTables();
-		$this->tableManager->migrate();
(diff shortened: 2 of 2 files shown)


notEvil:
No, because in the context of convert-type it is has to operate on a different database. During migration, the class gets instantiated only once, remembering the source schema (say SQLite) in the process. However, it then needs to create the tables on the target schema (say Postgres) using the target connection. So it needs to delegate the actual execution to Nextcloud.

dartcafe:
Thanks for contribution!!

dartcafe:
> Please note that I'm not particularly familiar with php and the codebase in general.

Fun fact. I am not even a programmer and had no clue of PHP or JS when I started contributing here .

notEvil:
Interesting :)

Regarding the failed check:
https://github.com/nextcloud/server/blob/e9ac0287f2b7d4da856bec416a282c8fc4c7feab/lib/private/DB/MigrationService.php#L444C3-L444C3
passes a `OC\DB\SchemaWrapper` but declares it as `OCP\DB\ISchemaWrapper`. And my quick and dirty solution was to not care about the type. This is where someone with more experience immediately knows what to do about it, closing the circle.

dartcafe:
Erks. I think, we need another approach. I will come back to it later. Currently I'm rather busy.

dartcafe:
Superseded by #3024

PR #3039, opened 2023-08-27
send all invitations

### LINKED ISSUE #1304: Send E-Mail to all "unsent invitations"
This is part 1/3 of the issue #1268

It is not possible to invite all not yet invited shares (unsent invitations).
I have contact groups with a lot of contacts. With the #902 I can now search for them and add them to the pool. But with the current implementation I have to first click on resolve to individual shares and then invite every contact individually.

A Button to send an invocation email to all shares which are not invited via email yet

image

dartcafe:
resolves #1304

* Resolve contact groups and circles, 
* send invitations to all shares, that did not have an invitation so far

also refactor/maintenance: 
 * unify sending results

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [5.3.0] - tbd
 ### New
  - Add label to public shares
+ - Send all unsent invitations to a poll with one click (resolves contact groups and circles too)
 ### Fixes
  - Fix API calls
 ### Changes
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -89,6 +89,7 @@ return [
 
 		['name' => 'share#list', 'url' => '/poll/{pollId}/shares', 'verb' => 'GET'],
 		['name' => 'share#add', 'url' => '/poll/{pollId}/share', 'verb' => 'POST'],
+		['name' => 'share#send_all_invitations', 'url' => '/poll/{pollId}/inviteAll', 'verb' => 'PUT'],
 		['name' => 'share#delete', 'url' => '/share/{token}', 'verb' => 'DELETE'],
 		['name' => 'share#personal', 'url' => '/share/personal', 'verb' => 'POST'],
 		['name' => 'share#sendInvitation', 'url' => '/share/{token}/invite', 'verb' => 'POST'],
--- a/lib/Command/Share/Add.php
+++ b/lib/Command/Share/Add.php
@@ -95,7 +95,7 @@ class Add extends Base {
 		foreach ($userIds as $userId) {
 			try {
 				$share = $this->shareService->add($poll->getId(), User::TYPE, $userId);
-				$this->shareService->sendInvitation($share->getToken());
+				$this->shareService->sendInvitation($share);
 			} catch (ShareAlreadyExistsException $e) {
 				// silently ignore already existing shares
 			}
@@ -110,7 +110,7 @@ class Add extends Base {
 		foreach ($groupIds as $groupId) {
 			try {
 				$share = $this->shareService->add($poll->getId(), Group::TYPE, $groupId);
-				$this->shareService->sendInvitation($share->getToken());
+				$this->shareService->sendInvitation($share);
 			} catch (ShareAlreadyExistsException $e) {
 				// silently ignore already existing shares
 			}
@@ -125,7 +125,7 @@ class Add extends Base {
 		foreach ($emails as $email) {
 			try {
 				$share = $this->shareService->add($poll->getId(), Email::TYPE, $email);
-				$this->shareService->sendInvitation($share->getToken());
+				$this->shareService->sendInvitation($share);
 			} catch (ShareAlreadyExistsException $e) {
 				// silently ignore already existing shares
 			}
--- a/lib/Command/Share/Remove.php
+++ b/lib/Command/Share/Remove.php
@@ -95,7 +95,7 @@ class Remove extends Base {
 	private function removeUsers(Poll $poll, array $userIds): void {
 		foreach ($this->getUserShares($poll) as $share) {
 			if (in_array($share->getUserId(), $userIds, true)) {
-				$this->shareService->delete($share->getToken());
+				$this->shareService->delete($share);
 			}
 		}
 	}
@@ -107,7 +107,7 @@ class Remove extends Base {
 	private function removeGroups(Poll $poll, array $groupIds): void {
 		foreach ($this->getGroupShares($poll) as $share) {
 			if (in_array($share->getUserId(), $groupIds, true)) {
-				$this->shareService->delete($share->getToken());
+				$this->shareService->delete($share);
 			}
 		}
 	}
@@ -119,7 +119,7 @@ class Remove extends Base {
 	private function removeEmails(Poll $poll, array $emails): void {
 		foreach ($this->getEmailShares($poll) as $share) {
 			if (in_array($share->getEmailAddress(), $emails, true)) {
-				$this->shareService->delete($share->getToken());
+				$this->shareService->delete($share);
 			}
 		}
 	}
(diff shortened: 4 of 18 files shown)


dartcafe:
@PhrozenByte Could you have a look at the tests. I changed some methods, but the tests fail now due to changed attributes and return values. Would be a big help.

PhrozenByte:
You changed the parameters of `ShareService::sendInvitation()`, just change the parameters in the tests as well.

Add `use (…, &$expectedShares)` (first define `$expectedShares` as empty array) to [this callback](https://github.com/nextcloud/polls/blob/7bc273989d7c74c2c307c13c6c66fcf8b4a49f49/tests/Integration/Command/Share/AddTest.php#L120-L129) and add the mocked `Share` objects (the result of `$this->createShareMock()`) to the `$expectedShares` variable, then change [this test](https://github.com/nextcloud/polls/blob/7bc273989d7c74c2c307c13c6c66fcf8b4a49f49/tests/Integration/Command/Share/AddTest.php#L134) to check for `$expectedShares` instead of `$expectedInvitationShareTokens`. You can then remove [this code](https://github.com/nextcloud/polls/blob/7bc273989d7c74c2c307c13c6c66fcf8b4a49f49/tests/Integration/Command/Share/AddTest.php#L103-L108). 

For `RemoveTest` it's even simpler, you replace [this code](https://github.com/nextcloud/polls/blob/7bc273989d7c74c2c307c13c6c66fcf8b4a49f49/tests/Integration/Command/Share/RemoveTest.php#L99C8-L100) by adding the mocked `Share` objects (again the result of `$this->createShareMock()`) to a new `$expectedShares` array and then use this variable for [this check](https://github.com/nextcloud/polls/blob/7bc273989d7c74c2c307c13c6c66fcf8b4a49f49/tests/Integration/Command/Share/RemoveTest.php#L119) instead.

This should do the trick. Unfortunately I don't have a dev setup for the polls app anymore.

PhrozenByte:
You changed the parameters of `ShareService::sendInvitation()`, just change the parameters in the tests as well.

Add `use (…, &$expectedShares)` (first define `$expectedShares` as empty array) to [this callback](https://github.com/nextcloud/polls/blob/7bc273989d7c74c2c307c13c6c66fcf8b4a49f49/tests/Integration/Command/Share/AddTest.php#L120-L129) and add the mocked `Share` objects (the result of `$this->createShareMock()`) to the `$expectedShares` variable, then change [this test](https://github.com/nextcloud/polls/blob/7bc273989d7c74c2c307c13c6c66fcf8b4a49f49/tests/Integration/Command/Share/AddTest.php#L134) to check for `$expectedShares` instead of `$expectedInvitationShareTokens`. You can then remove [this code](https://github.com/nextcloud/polls/blob/7bc273989d7c74c2c307c13c6c66fcf8b4a49f49/tests/Integration/Command/Share/AddTest.php#L103-L108). 

For `RemoveTest` it's even simpler, you replace [this code](https://github.com/nextcloud/polls/blob/7bc273989d7c74c2c307c13c6c66fcf8b4a49f49/tests/Integration/Command/Share/RemoveTest.php#L99C8-L100) by adding the mocked `Share` objects (again the result of `$this->createShareMock()`) to a new `$expectedShares` array and then use this variable for [this check](https://github.com/nextcloud/polls/blob/7bc273989d7c74c2c307c13c6c66fcf8b4a49f49/tests/Integration/Command/Share/RemoveTest.php#L119) instead.

This should do the trick. Unfortunately I don't have a dev setup for the polls app anymore.

dartcafe:
Something I thught regarding the shareMocks, but I got stuck while trying to understand it. Thanks. I give it a try. I am not familiar with PHPUnit, so... Maybe I will learn something. ;-) 

The main problem is, I can't get it to work with my Windows machine, which is my default dev env.

dartcafe:
@PhrozenByte I suspect the expected return is no more an array, but an object (of `OCA\Polls\Model\SentResult`). How can I fix the test to match the expected value (or better define the object to be expected). 

Did I understand your suggestions right (see last commit)?

PhrozenByte:
Ah, sorry, my bad, I missed that this logic can't work: Our mocked `ShareService:add()` (lines 116-127) is called *after* we set our expectations for `ShareService::sendInvitation()` (line 129-132), thus `$expectedShares` is empty.

We can't rely on `ShareService:add()` to create the mocked `Share` objects for us, we must create them ourselves first, like the following:

```php
$mockedShares = [];
$expectedShares = [];
foreach ($pollData['expectedInvitations'] ?? [] as $type => $shares) {
	foreach ($shares as $userId) {
		if (!in_array($userId, $pollData['initialShares'][$type] ?? [])) {
			$share = $this->createShareMock($pollData['pollId'], $type, $userId);
			$mockedShares[$type][$userId] = $share;
			$expectedShares[] = $share;
		}
	}
}
```

And then change our mocked `ShareService:add()` to return the matching mocked `Share` object from `$mockedShares` instead:

```php
			->willReturnCallback(function (int $pollId, string $type, string $userId = '') use ($pollData, $mockedShares): Share {
				$userIdConstraint = $this->logicalOr(...$pollData['expectedShares'][$type] ?? []);
				$userIdConstraint->evaluate($userId);

				if (in_array($userId, $pollData['initialShares'][$type] ?? [])) {
					throw new ShareAlreadyExistsException();
				}

				return $mockedShares[$type][$userId];
			});
```

The suggested changes for `RemoveTest` should work.

dartcafe:
Hey. Thanks. Helped me for understanding these tests. I adoped your change, see, if it runs through now. 👍

dartcafe:
@PhrozenByte I am not sure, if I just created an always true condition. But I changed the input parameters to `$initialShares` and the test runs through. Does it look valid?

PhrozenByte:
Switching the `RemoveTest` to `$initialShares` would indeed always yield true, because you'd allow any of the existing shares to be removed, not just the ones you actually want to remove (consider we have shares ABCD and you want to remove shares AD, using `$initialShares` accepts any of ABCD, not just AD). You need the following:

```php
		$initialShares = [];
		$expectedShares = [];
		foreach ($pollData['initialShares'] ?? [] as $type => $shares) {
			foreach ($shares as $userId) {
				$share = $this->createShareMock($pollData['pollId'], $type, $userId);
				$initialShares[] = $share;

				if (in_array($userId, $pollData['expectedShares'][$type] ?? [])) {
					$expectedShares[] = $share;
				}
			}
		}
```

Replace `$expectedShareCount` by `count($expectedShares)` and `$expectedShareTokens` by `expectedShares` below.

On a side note, our tests (both in `RemoveTest`, nor `AddTest` btw) aren't perfect. Right now we only test whether the shares passed to `ShareService::add()` and `ShareService::delete()` match *any* of the expected shares (`->with($this->logicalOr(...$expectedShares))` in `RemoveTest`) and that the number of calls is correct (`->expects($this->exactly(count($expectedShares)))` in `RemoveTest`). Thus the test would still pass if one share is passed twice and another share is left out. I don't believe that this is a big issue, but if you want to improve it and actually test whether all expected shares are passed exactly once, you'd need to write a custom check function with `->with(self::callback(function (Share $share): bool { /* body */ }))`.

dartcafe:
Argh I was confused, but was the source by myself. I pushed $userId to the $expectedShares array instead of the mockedShare here: https://github.com/nextcloud/polls/pull/3039/commits/53948c9d5b9030ac4a63732e121528e765f96d2b. 🙈

dartcafe:
> On a side note, our tests (both in `RemoveTest`, nor `AddTest` btw) aren't perfect.

At least there are tests. 😆 The rest of the project would need a full test suite, but I have no time for that. Since I know the structure of this app, I can test most of it by making manual tests. But one day ... who knows 😉

PhrozenByte:
Nice, tests pass now. Great work, LGTM! :+1:

> At least there are tests. 😆 The rest of the project would need a full test suite, but I have no time for that. Since I know the structure of this app, I can test most of it manually. But one day ... who knows 😉

I know the struggle :see_no_evil: Creating tests isn't my favourite job either. To be honest, if it's mostly a single developer project and this developer has a complete overview over the whole project (like you do), and if the project is no library that is used by other developers (like this app), one can argue that tests don't provide a major advantage. Sure, they undoubtedly improve things and are much appreciated, but writing them takes a lot of effort and time, which is often better spent in developing new features. This only changes - but then drastically - the more developers are working on the project. But if that's not the case I don't see much of a problem when there is no test suite...

dartcafe:
Thanks again. I'll merge for the new beta

Issue #2855, opened 2023-04-13
Public Shared Polls queries for users in a loop?


b90g:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).
- [X] This issue is **not** already reported on Github _(I've searched it)_.

### What went wrong, what did you observe?

When i share [a poll publicly](https://wolke.dev.netzbegruenung.de/apps/polls/s/P3gbfNAX2Y9oZwgr) and enter a user name the interfaces queries for existing user names until it breaks. in my browser meanwhile i get a green check mark that i can use that user name. when i submit my user name to start polling the  dialog windows stays afloat the poll and i can not enter my polling options. 

### What did you expect, how polls should behave instead?

1. I open the link.
2. I enter a name so people know who submitted their  poll options
3. I submit my polls
fin

### What steps does it need to replay this bug?

1.Publicly share a poll (see first paragraph)

### Installation method

Installed/updated from the appstore (Apps section of your site)

### Installation type

First time installation

### Affected polls version

5.0.0

### Which browser did you use, when experiencing the bug?

- [X] Firefox
- [ ] Chrome
- [X] Chromium/Chromium based (i.e. Edge)
- [ ] Safari
- [ ] Other/Don't know

### Other browser

(112 versions on both browsers, no plugins installed)

### Which System did you use, when experiencing the bug?

- [ ] Windows
- [X] Linux
- [ ] MacOs
- [ ] iOS
- [ ] Android
- [ ] Other/Don't know

### Add your browser log here

```shell
$ is deprecated: The global jQuery is deprecated. It will be removed in a later versions without another warning. Please ship your own.
```

### NC version

Nextcloud 25 (Nextcloud Hub 3)

### PHP engine version

PHP 8.1

### Database engine

MariaDB

### Database Engine version or other Database

10.5.18

### Which user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [X] SSO - SAML
- [ ] Other/Don't know

### Add your nextcloud server log here

```shell
after entering the user name:


[...]

PR #3091, opened 2023-09-27
improve validatePublicUsername() performance

JonathanTreffler:
On Instances with many groups the validatePublicUsername endpoint ends with a timeout, because the code checking if the anonymous users chosen userName matches any group on the server is quite inefficient. It does not utilize the search parameter of the Group::search() function and therefore loads all groups from all group backends from the database into RAM, creates a php entity for every one of them and calls two functions on every one of those entities.

I created a performance profile of a simple call of the /check/username endpoint on an instance with ~12k groups (which is certainly a lot, but I assure you every one of those has an important role on our instance and other apps have no problems working with this amount of groups):

In total this request took 71.6 seconds, causing the connection to the client to drop during the request with a timeout (the php process still runs to completion).

In the performance profile you can see, that most time (73%) is spent in the nextcloud core just fetching the data from the database and putting it into php entities.

image

In total this one request causes 54 THOUSAND requests to QueryBuilder->execute(), which AFAIK is below the cache layer and that means every one of those is a sql query to  the database. I hope we can all agree that kind of amplification from one request is not great.

But the current implementation makes sense looking at the timeline: The first usage of this empty search parameter I could track down through various refactors is the initial implementation of Public Pages back in 2019 in #664 .
This was before the default database group backend started searching by display names in 2020 (introduced in https://github.com/nextcloud/server/pull/21358), so it makes sense a custom search was implemented on this high level. But now we can use the MUCH faster filtering on the database query level :partying_face: .

It is still true, that every group backend decides for itself, how it handles the search parameter and theoretically a group backend could decide only to search by gid, but I think it is reasonable to assume, that any group backend, that uses DisplayNames (ours with the 12k groups does not) follows the core implementation and also searches display names.

On our instance this small 9 characters of code patch brought the request time of the /check/username endpoint to 714 ms down from 71.6 seconds, a 10200% performance improvement :partying_face: .
image

### DIFF
--- a/lib/Service/SystemService.php
+++ b/lib/Service/SystemService.php
@@ -145,8 +145,9 @@ class SystemService {
 			throw new InvalidUsernameException;
 		}
 
-		// get all groups
-		foreach (Group::search() as $group) {
+		// get all groups, that include the requested username in their gid
+		// or displayname and check if any match completely
+		foreach (Group::search($userName) as $group) {
 			if ($userName === strtolower(trim($group->getId()))
 				|| $userName === strtolower(trim($group->getDisplayName()))) {
 				throw new InvalidUsernameException;


dartcafe:
First of all, thanks for your excellent analysis. 

12k groups. Wow. Never thought that this is a valid use case. Your suggestion will speed up it for sure, but it will eliminate the thought of avoiding spoofing by using exiting displaynames. 

But I have to rethink about @svenseeberg's [comment](https://github.com/nextcloud/polls/issues/2855#issuecomment-1711935980). Maybe it would be better to remove the check against group names, because groups can't be a participants. The check as it is will probably lose its value.

To be honest: I have to dig in my memory why we introduced it back then.

dartcafe:
Ah, wait. Reading helps. I did not get the fact, that the search also includes the search for the group's displayname. In this case, adding the serach term to the group search will improve it indeed. 

Nevertheless, we have to rethink the check.

dartcafe:
Tracking this down a little bit further: Since the IGroup:search() does an exact case insensitive search, it would be enough to simply check (`OCA\Polls\Model\Group\Group:exists(query)`), if there is a return from the search.

This way the object creation could be skipped also.

dartcafe (approved):
Since this should work...

dartcafe:
Please merge by yourself, unless you plan further improvements.

JonathanTreffler:
> Tracking this down a little bit further: Since the IGroup:search() does an exact case insensitive search

Not exactly, IGroup:search() on the default database group backend does a [SQL iLike](https://www.geeksforgeeks.org/postgresql-ilike-operator/) query with `'%' . $this->dbConn->escapeLikeParameter($search) . '%'` (see https://github.com/nextcloud/server/blob/a636aef585a6e3e6cc628be6952cbec526148d70/lib/private/Group/Database.php#L291).
Meaning the returned groups only have to contain the search string anywhere in the gid or display Name.

That is why I left in the for loop checking the returned entries, because both exact match and partial matches will get returned from the search.

JonathanTreffler:
> Please merge by yourself, unless you plan further improvements.

If we detect other issues in the roll out of the app onto our huge instance we will probably be back for other optimizations :) , but for now I think this can be merged.

dartcafe:
Sure. All contributions are welcome here. Feeling less lonly. 😆

svenseeberg:
Thank you all :)

PR #3109, opened 2023-10-15
[5.4.0-beta5] Changed revoking to locking

dartcafe:
* The terms have been changed from `revoked` to `locked`. It is a simple and precise term. 
  `disabled` clashes with disabled states of components in the code and would cause more work.
  `activated` would possibly need more information, because there are no actions and could be misinterpreted.
* The database schema and the usage of the term `locked `is used consistently in the code instead of `revoked`
* Locking/unlocking is now possible for all shares, no matter if it is a public share or if the user has voted or not
* The lock action has been moved to the context menu of the share and deletion of shares has returned to all shares.
* Users of locked shares get an information banner on the top of the poll page, that tey just have read access.

Revoking of a share instead of deleting will come back in another pr to replace the deletion counter, which is ugly IMHO.

lock share | unlock share | unlocked shares list | user information
-|-|-|-
grafik | grafik | grafik | grafik


PhrozenByte:
As I already said in https://github.com/nextcloud/polls/issues/3107#issuecomment-1763355252, looks great, thank you! :+1:

Just some quick question: Do I read the code correctly that it will just drop the `revoked` column and add the new `locked` column on update from beta 3 or 4, i.e. I don't have to take care of anything and just mind that all previously revoked shares will be unlocked, correct? I'm running a poll with beta 4 in production till tomorrow, just so that I don't accidentally break anything. I'll give some feedback about the new feature tomorrow when the poll closes.

dartcafe:
> I don't have to take care of anything and just mind that all previously revoked shares will be unlocked, correct?

That is the intention and how it works here. beta6 will come later the day.

PR #3148, opened 2023-11-08
Not null violation error in migrations

fsamapoor:
A bug has been introduced with https://github.com/nextcloud/polls/commit/3da60e4457c5100ccf9184a721a16317bffe1447 which is causing migration errors when you try to upgrade the app.

```
Database error when running migration 050100Date20230515083001 for app polls

An exception occurred while executing a query: SQLSTATE[23502]: Not null violation: 7 ERROR:  column "session_id" contains null values
```

Having a non-nullable column with the default value of `null` is definitely not right. likely introduced by mistake with the refactoring of the class.

In this PR, I have made the `session_id` column nullable which is the intended behavior.

I have also reviewed all other combinations on the file and found no other non-nullable columns with a default value of `null`.

This is not being reflected with the new installations of the app, as the database handles the error seamlessly. However, as mentioned earlier, an error is thrown when upgrading older versions of the app.

### DIFF
--- a/lib/Migration/TableSchema.php
+++ b/lib/Migration/TableSchema.php
@@ -242,7 +242,7 @@ abstract class TableSchema {
 			'poll_id' => ['type' => Types::BIGINT, 'options' => ['notnull' => true, 'default' => 0, 'length' => 20]],
 			'table' => ['type' => Types::STRING, 'options' => ['notnull' => true, 'default' => '', 'length' => 64]],
 			'updated' => ['type' => Types::BIGINT, 'options' => ['notnull' => true, 'default' => 0, 'length' => 20]],
-			'session_id' => ['type' => Types::STRING, 'options' => ['notnull' => true, 'default' => null]],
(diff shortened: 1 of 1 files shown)


dartcafe:
Yes. right, the definition is wrong. But what confused me, is that there was an error triggered. The table gets dropped and recrated with updates. Unfortunately you did not provide an issue with your system setup and the triggered error message.

dartcafe (approved):

fsamapoor:
> Yes. right, the definition is wrong. But what confused me, is that there was an error triggered. The table gets dropped and recrated with updates. Unfortunately you did not provide an issue with your system setup and the triggered error message
> 
> **Edit:** ok. I can interprate the location of the error. But there should no error message, because the table should have no entiries.

I don't have direct access to the Nextcloud instance our deployment team attempted to upgrade. I know it was a v24 Nextcloud server using Postgres, and the plan was to upgrade it to v26. I'll provide the necessary information in a new issue once I gain access to the server.

dartcafe:
Yes. Would be good. Possibly there is a migration path, I did not take into account.

Issue #3280, opened 2024-01-30
Data loss going from v3 to v5


ChristophWurst:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### What went wrong, what did you observe?

As the admin of a production setup I'm going through an upgrade marathon. The Nextcloud setup is upgraded step by step from 22 to 27, Polls from 3.8.4 to 5.4.2.
We have noticed that data is lost once the migration is done. I've analyzed it and it seems to be caused by a wrong order of migration and repair steps.
1) The two migrations are executed and add columns like polls_votes.vote_option_hash
2) `\OCA\Polls\Db\TableManager::deleteAllDuplicates` runs ``SELECT DISTINCT `t1`.`id` FROM `*PREFIX*polls_votes` `t1` INNER JOIN `*PREFIX*polls_votes` `t2` ON `t1`.`id` < `t2`.`id` WHERE (`t1`.`poll_id` = `t2`.`poll_id`) AND (`t1`.`user_id` = `t2`.`user_id`) AND (`t1`.`vote_option_hash` = `t2`.`vote_option_hash`)``. At this point, vote_option_hash is an empty string. Therefore the full table is wrongly considered duplicates and data is DELETEd.
3) `\OCA\Polls\Db\TableManager::migrateOptionsToHash` runs to fill the columns that have just been deleted.

### What did you expect, how polls should behave instead?

1) Upgrade schema
2) Fill hash columns
3) Deduplicate

### What steps does it need to replay this bug?

1. Set up v3
2. Fill polls+votes tables
3. Switch to v5
4. Run occ upgrade

With adjustment of nextcloud-max in info.xml I was able to replay the data loss on my server master/29 dev setup.

### Installation method

Installed/updated from the appstore (Apps section of your site)

### Installation type

Updated from previous major version (i.e. 3.x.x to 4.x.x)

### Affected polls version

v5.2.2

### Which browser did you use, when experiencing the bug?

- [ ] Firefox
- [ ] Chrome
- [ ] Chromium/Chromium based (i.e. Edge)
- [ ] Safari
- [X] Other/Don't know

### NC version

Nextcloud 27 (Nextcloud Hub 5)

### PHP engine version

Other/Don't know

[...]

PR #3281, opened 2024-01-30
create hashes before deleting invalid entries

dartcafe:
related #3280

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -63,8 +63,8 @@
     <post-migration>
       <step>OCA\Polls\Migration\RepairSteps\DropOrphanedTables</step>
       <step>OCA\Polls\Migration\RepairSteps\DropOrphanedColumns</step>
-      <step>OCA\Polls\Migration\RepairSteps\DeleteInvalidRecords</step>
       <step>OCA\Polls\Migration\RepairSteps\UpdateHashes</step>
+      <step>OCA\Polls\Migration\RepairSteps\DeleteInvalidRecords</step>
       <step>OCA\Polls\Migration\RepairSteps\CreateIndices</step>
     </post-migration>
     <install>


ChristophWurst (approved):
Didn't test but makes sense if Nextcloud keeps the order

dartcafe:
I rely on it, otherwise the repair steps would not be predictable.

lookup82:
Unfortunately, I have performed the update and cannot roll it back. Can I retrieve the surveys? I have backups available.
From the administration menu, some surveys can be seen; could this be useful?

dartcafe:
It is the options and votes, which got deleted. You can rollback from a db backup and then call occ polls:db:rebuild. 

The command works in the correct order, other than the repair steps and should create the hashes.

dartcafe:
@v1r0x Are you able to update the version in the app store, before more users run into the problem?

lookup82:
> It is the options and votes, which got deleted, so the polls themself are still there. You can rollback from a db backup (tables `oc_options` and `oc_votes` should be enough and then call occ polls:db:rebuild.
> 
> The command works in the correct order, other than the repair steps and should create the hashes.

Performing these operations does not seem adequate; I am investigating further.

mysqldump -u nextcloud -p nextcloud oc_polls_options oc_polls_votes  > backup_polls.sql
mysqldump -u nextcloud -p nextcloud < backup_polls.sql
occ polls:db:rebuild

lookup82:
I am not sure if `oc_options` and `oc_votes` were empty even before performing the restore from the old database.

Nonetheless, I encountered the following error in the log:
`json_decode(): Argument #1 ($json) must be of type string, null given in file '/var/www/nextcloud/apps/polls/lib/Db/Preferences.php' line 71`

I resolved the issue following a suggestion I found in this comment: https://github.com/nextcloud/polls/issues/3283#issuecomment-1917905262

Thanks to all 🙏🏻

Issue #2926, opened 2023-05-30
Migration to Vue 3


dartcafe:
# Migrate to Vue 3

## Ressources
* [Vue Migration Guide](https://vuex.vuejs.org/guide/migrating-to-4-0-from-3-x.html)
* [Vuex Migration Guide](https://vuex.vuejs.org/guide/migrating-to-4-0-from-3-x.html)
* [Vue Router Migration Guide](https://router.vuejs.org/guide/migration/)
* [Vuex to Pinea Migration](https://pinia.vuejs.org/cookbook/migration-vuex.html)

## Blocking Dependencies
* @nextcloud-libraries/nextcloud-vue
  * https://github.com/nextcloud-libraries/nextcloud-vue/issues/5950

## To dos:

- [x] https://github.com/nextcloud/polls/issues/3758
- [ ] Wait for https://github.com/nextcloud-libraries/nextcloud-vue/milestone/118

dartcafe:
done. 🎉

PR #3311, opened 2024-02-12
Migration to Vue 3

dartcafe:
relates #2926


dartcafe:
outdatend and no more needed

Issue #3345, opened 2024-02-25
Existing votes no longer displayed


Findus23:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### What went wrong, what did you observe?

I recently updated polls to 6.1.3 (I unfortunately don't know the previous version, but as the last update was a few weeks ago I assume it was a 6.0.x version).

Since then all my polls show that nobody voted on it. There are no errors in the JS console related to it and I don't think it is a frontend issue as `https://my.nextcloud.example/apps/polls/poll/153/votes` returns `{"votes":[]}` for all poll IDs.
This affects both the votes done by me (logged in user, creator of the poll) and everyone else (guest users from shared link).

The `oc_polls_votes` still contains all data, so I don't think the votes actually got lost. 

```
MariaDB root@(none):nextcloud> select * from oc_polls_votes order by id desc limit 2
+------+---------+-------------+----------------+-------------------------------------------------------+-------------+----------------------------------+---------+
| id   | poll_id | user_id     | vote_option_id | vote_option_text                                      | vote_answer | vote_option_hash                 | deleted |
|------+---------+-------------+----------------+-------------------------------------------------------+-------------+----------------------------------+---------|
| 6352 | 153     | ex_GuArCDsu | 1354           | 2024-02-24T23:00:00+00:00 - 2024-02-25T23:00:00+00:00 | yes         | 3c4d161570dec3157e794f8077788a08 | 0       |
| 6351 | 153     | ex_GuArCDsu | 1350           | 2024-02-20T23:00:00+00:00 - 2024-02-21T23:00:00+00:00 | yes         | e201cf3708b7d17867e81c5e70ab13ae | 0       |
+------+---------+-------------+----------------+-------------------------------------------------------+-------------+----------------------------------+---------+
```

If I try to vote again, the `PUT` to `https://my.nextcloud.example/apps/polls/vote` fails with a 500 

```text
[index] Fehler: An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '153-Lukas-480ff97bc90fecfcbc30825b61074017' for key 'UNIQ_votes'
	PUT /apps/polls/vote
[...]

PR #3346, opened 2024-02-25
update issue template for newer PHP and Nextcloud versions

Findus23:
When creating #3345 I noticed that some PHP and Nextcloud options can't be picked as an option in the issue template


dartcafe (approved):
Nice. Thank you!

Issue #3362, opened 2024-03-14
N+1 problem in `\OCA\Polls\Service\PollService::list`


ChristophWurst:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### What went wrong, what did you observe?

I'm analyzing a production installation of this app and noticed that the app loads data in a loop. The more polls you have, the more database queries run. This leads to a scaling issue on large installations.

From the query log:

```sql
SELECT * FROM `oc_polls_polls` WHERE (`deleted` = :dcValue1) OR (`owner` = :dcValue2)
# Poll 1
SELECT * FROM `oc_polls_share` WHERE (`poll_id` = :dcValue1) AND (`user_id` = :dcValue2)
SELECT `timestamp` FROM `oc_polls_options` WHERE `poll_id` = :dcValue1
SELECT * FROM `oc_polls_votes` WHERE (`poll_id` = :dcValue1) AND (`user_id` = :dcValue2)
# Poll 2
SELECT * FROM `oc_polls_share` WHERE (`poll_id` = :dcValue1) AND (`user_id` = :dcValue2)
SELECT `timestamp` FROM `oc_polls_options` WHERE `poll_id` = :dcValue1
SELECT * FROM `oc_polls_votes` WHERE (`poll_id` = :dcValue1) AND (`user_id` = :dcValue2)
```

^ the first query has to happen. The other blocks are repeated for every row returned in the first query.

### What did you expect, how polls should behave instead?

Either join all tables and load data in one query, or at least combine the queries for child data with a `WHERE IN` for all loaded polls.

### What steps does it need to replay this bug?

1. Install the app
2. Add two polls
3. Enable query logging
4. Load the polls start page

### Installation method

Installed/updated from the appstore (Apps section of your site)

### Installation type

First time installation

### Affected polls version

5.4.3

### Which browser did you use, when experiencing the bug?

- [ ] Firefox
- [ ] Chrome
- [ ] Chromium/Chromium based (i.e. Edge)
- [ ] Safari
- [ ] Other/Don't know

### NC version

Nextcloud 26

### PHP engine version

PHP 8.0

### Database engine

MySQL

### Which user-backends are you using?

- [ ] Default user-backend _(database)_
[...]

PR #3364, opened 2024-03-14
perf: Determine participant status more efficiently

ChristophWurst:
This makes one of the N+1 queries of https://github.com/nextcloud/polls/issues/3362 more efficient.

Before: load all votes of the poll, create entity objects and then do a count
After: count in the database


dartcafe:
closing in favour of other approches

PR #3368, opened 2024-03-15
perf: Cache \OCA\Polls\Model\Acl::getIsInvolved

ChristophWurst:
image

Two polls with one share cause 32 requests for the share recipient. Caching the result of `\OCA\Polls\Model\Acl::getIsInvolved` in memory cuts these to 18 requests.


dartcafe:
closing in favour of other approches

Issue #3147, opened 2023-11-07
Help wanted for writing PHP tests


dartcafe:
## We are hiring a developer for writing tests for the PHP backend. 

Top conditions: 
* 100% remote work
* Free fruits and drinks in your refrigerator (if you buy then)
* Self-determinded working times
* ...

Or in other words: Who is able and willing to write tests and support this app?

jovialcore:
Hi. I am willing to write the tests. 
Is this position still available?

dartcafe:
> Hi. I am willing to write the tests.

You would make me happy with contibuting and maintaining tests!

PR #3373, opened 2024-03-16
join votes for participation check

dartcafe:
Count current user's votes via join in poll
regarding #3362

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -175,6 +175,7 @@ class Application extends App implements IBootstrap {
 		$context->registerService(PollMapper::class, function (ContainerInterface $c): PollMapper {
 			return new PollMapper(
 				$c->get(IDBConnection::class),
+				$c->get(UserMapper::class)
 			);
 		});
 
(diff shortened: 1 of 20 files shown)


lib/AppInfo/Application.php
```diff
@@ -132,4 +149,66 @@ public function register(IRegistrationContext $context): void {
...
+	private function registerServices(IRegistrationContext $context) {
+		$context->registerService(UserMapper::class, function (ContainerInterface $c): UserMapper {
+			return new UserMapper(
+				$c->get(IDBConnection::class),
+				$c->get(ISession::class),
+				$c->get(IUserSession::class),
+				$c->get(IUserManager::class),
+				$c->get(LoggerInterface::class),
+			);
+		});
```
come-nc, line 165: Is all this needed? This is the default DI behavior to my knowledge.

dartcafe: I was annoyed by the unit tests, where every deopendency needed to be instatiated. 
If there is a DI possibility for the tests, this is indeed unnecessary.

come-nc: Usually you use Mocks for the tests

dartcafe: Yeah. Tests. See https://github.com/nextcloud/polls/issues/3147 😉

PR #3445, opened 2024-04-23
Perf improvements for stable-6 for 26 branch

come-nc:
Trying to bring the number of DB requests for polls listing down.
So far got down from 52 to 30 for 4 polls with no votes.

### DIFF
--- a/lib/Db/UserMapper.php
+++ b/lib/Db/UserMapper.php
@@ -127,13 +127,16 @@ class UserMapper extends QBMapper {
 			// just catch and continue if not found and try to find user by share;
 		}
 
-		try {
-			$share = $this->getShareByPollAndUser($userId, $pollId);
-			return $this->getUserFromShare($share);
-		} catch (ShareNotFoundException $e) {
-			// User seems to be probaly deleted, use fake share
-			return new Ghost($userId);
+		if ($pollId !== null) {
+			try {
+				$share = $this->getShareByPollAndUser($userId, $pollId);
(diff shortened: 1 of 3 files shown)


artonge (approved):

sorbaugh:
Hello dartcafe, I'd also like to apologize for making changes without your review and coordination. 🙏 To be transparent, we are facing pressure and urgency due to a support case with a customer that heavily uses Polls and is encountering some performance issues.

In the spirit of not bypassing your role as the maintainer and continue on a more collaborative way, the changes we made are:
- Make the My Polls app the default view
- Add a spinner as feedback to the user for the loading time
- Attempt to optimize the query for My Polls (in our test the loading time was reduced from 50ish seconds to 30) This was the PR: https://github.com/nextcloud/polls/pull/3445

If you have input on what could be done to bring the time down to, say, 10 seconds, your input would be invaluable :)

dartcafe:
> /backport to master

@artonge Never ever!!!!

come-nc:
> > /backport to master
> 
> @artonge Never ever!!!!

Is the code base that different? We should make sure master does not suffer from the perf problems this PR fixed for 6, otherwise upgrading from 6 to 7 would result in performance degradation. I did not look into the details, maybe those extra DB requests are already gone in 7.

dartcafe:
1. Yes, the code differs a lot
2. Most of the optimizations here come from the master branch

Existing problems have to be solved another way.

PR #3450, opened 2024-04-24
Default to "My polls" view

artonge:


ArtificialOwl (approved):

dartcafe:
@artonge No. Please revert this PR. The "relevant view" is default by intention.

artonge:
The idea was to avoid the long loading time in case the user as access to a lot of polls.
Maybe this can be a good default until we properly tackle the performance issue?

dartcafe:
Not a good idea. It affects the user experience. 
* Users who rely on the default to the relevant polls will be confronted with a forced change
* The value of the change is not considered to all users
* Alternative approaches are not discussed (i.e. revert to the last used category)

Regarding the performance issue. I doubt, if the performance issue is solved this way. It maybe hidden after entering the app but comes back, when changing the category. If there is a performance issue, it should be analyzed first and the the solution can be decided. 

I am still not sure if it is 
* a render problem, 
* a problem caused by the store or 
* a filtering problem.

PR #3451, opened 2024-04-24
Default to "My polls" view

artonge:
Signed-off-by: Louis Chemineau <louis@chmn.me>


dartcafe:
@susnux @artonge 

Where comes this change request from? Why is this merged without any review from me or without any consideration about the value or alternatives?

See https://github.com/nextcloud/polls/pull/3450#issuecomment-2075506101

Please revert this PR, too.

PR #3455, opened 2024-04-25
Improve API documentation

Bagoult:
Hello! I improved the documentation of the REST-API to make it easier for future users. Here is what it does:

* Refreshes outdated information (`deleted` field is now a boolean, `share_api#add` endpoint was changed recently).
* Completes sections about payloads, and creates separate sections to give example of possible return values. (This way, the reader knows which fields appear in the return value, but shouldn't appear in the payload.)
* Explains a few things that I found confusing the first time I came across them (ex: sending `"expire": 0` doesn't remove the current expiration date, but sets it to 0).
* Another curl example to show how to create a new poll.


docs/API_v1.0.md
```diff
@@ -1,174 +1,408 @@
...
+```json
+{
+    "type": "datePoll",
+    "title": "Poll Title"
+}
+```
+
+### Update poll
+Send the full or a partial structure.
+In this example, we set an expiration date on January 1, 1970 at 00:00:01 UTC.
```
dartcafe, line 51: Additionally it could be mentioned, that an expiry timestamp in the past represents a closed poll, 0 represents an endless poll (aka without expiration) and an expiry timestamp in the future sets an automatic closing date. So this example closes the poll.

dartcafe:
Awesome. Great work. Unfortunately the structures are still subject of changing. Especially the update poll method needs some more work, since I plan to change the structure and differentiate between read only properties and properties wich can be changed.

dartcafe (changes requested):

dartcafe:
And, it would be good if you could sign your commits.

dartcafe:
Give me a hint, when you are finished. And thanks a lot, there were really a lot of missing things and broken docs.

dartcafe (approved):
Great so far.

Bagoult:
Everything should be signed off now and ready to merge @dartcafe :)

dartcafe:
@Bagoult Ah. And a rebase is not really necessary until you don't need the current target branch for testing your build.

Issue #3439, opened 2024-04-22
Optimize the query in the "My polls" section to query only for the polls of the user


sorbaugh:
### ⚠️ This issue respects the following points: ⚠️

- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Describe the goal you'd like to achieve

Currently, there seems to be performance issues when opening the "Relevant"/default view if there are lots and lots of polls in the database. It takes a long time to display information in this view, even if the user only has 2 polls. It seems that at the moment everything is loaded in the background despite it displaying only a few). 

### Describe possible solutions

* The "My Polls" view should be made the initially displayed view
* The "My Polls" view should only load those, and not filter for them afterwards. (dedicated query used, so the data can be loaded in an optimized, quicker way)

dartcafe:
What do you call a lot?

dartcafe:
Answer to https://github.com/nextcloud/polls/pull/3444#issuecomment-2076699146
> We are facing pressure and urgency due to a support case with a customer that heavily uses Polls and is encountering some performance issues.
> 

### My statements to the changes: 
> In the spirit of not bypassing your role as the maintainer and continue on a more collaborative way, the changes we made are:
>     * Make the My Polls app the default view

This affects all users because of the need of a single request. This is not how public software should be developed and we cannot act like this. This could raise possibly a lot of discussion and changes, reverts, changes and so on. This is why I rejected the change. But of course we can can discuss it in a specific, separate issue, what goal is to achieve.

There is a reason, why the template for change requests ask for a user story. 

> * Add a spinner as feedback to the user for the loading time

This seems to me like a bug, because the spinner was missing but should be there. So needs a fix and is the reason, why I accepted this PR. Although I already fixed it in a local branch, but to be fair, nobody could know that.

> * Attempt to optimize the query for My Polls (in our test the loading time was reduced from 50ish seconds to 30) This was the PR: [[stable6] Perf improvements for stable-6 for 26 branch #3445](https://github.com/nextcloud/polls/pull/3445)

This issue is under maintenance of the nextcloud team and only for the stable-6 branch. The master branch (v7.x) followed another approach to reduce loading times.

> If you have input on what could be done to bring the time down to, say, 10 seconds, your input would be invaluable :)

[...]

PR #3461, opened 2024-04-29
Allow to remove the poll list from the navigation by admin setting

dartcafe:
Allow deacivating loading polls inside the navigation out of performance reasons

This PR may help to increase the performance of loading polls referring [this observations](https://github.com/nextcloud/polls/issues/3439#issuecomment-2080454170), by adding a setting for diabloing loading the polls list into the navigation. 

@come-nc I separated this changes from the main PR, so it should be possible to backport them to the stable-6 branch.


src/js/views/Navigation.vue
```diff
@@ -131,6 +131,7 @@ export default {
 		...mapState({
 			isPollCreationAllowed: (state) => state.polls.meta.permissions.pollCreationAllowed,
```
dartcafe, line 132: @come-nc 
Watch the changed structure. For the stable-6 branch, the mapping should not chbe changed. (`meta.*` does not exist there in the store.)

PR #3474, opened 2024-05-02
WIP - Use a count instead of queries to get yes vote count

come-nc:
Not working yet, I get 4 instead of 2 with my test data, I think this is because orphaned votes should be excluded.

### DIFF
--- a/lib/Db/Poll.php
+++ b/lib/Db/Poll.php
@@ -136,6 +136,7 @@ class Poll extends EntityWithUser implements JsonSerializable {
 	protected int $maxDate = 0;
 	protected int $minDate = 0;
 	protected int $currentUserVotes = 0;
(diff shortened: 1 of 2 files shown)


dartcafe:
Right. Orphaned votes are identified by having no matching option.

dartcafe:
The problem ist the joined votes after joined options. So the counted votes get multiplied with the count of options.

Currently the counted votes are wrong, because of that. 

I tried to change the join to a subquery, but it behaves strange. see https://github.com/nextcloud/polls/pull/3477

dartcafe:
@come-nc Check the strategy in https://github.com/nextcloud/polls/pull/3477
BTW: Orphaned votes are correctly counted, since they block any participants vote limit, too. The affected user is informed about this and may delete theese orphaned votes.

PR #3477, opened 2024-05-03
Count user's votes by subquery

dartcafe:
### The current approach to count a user's votes by a join statement returns a by the count of options multiplied result

* Replace joipn by a subquery

### DIFF
--- a/lib/Db/Poll.php
+++ b/lib/Db/Poll.php
@@ -151,6 +151,8 @@ class Poll extends EntityWithUser implements JsonSerializable {
 	protected int $currentUserVotes = 0;
 	protected string $userRole = self::ROLE_NONE;
 	protected ?int $isCurrentUserLocked = 0;
+	protected ?int $currentUserVotesSub = 0;
+	protected ?int $currentUserVotesYesSub = 0;
 
 	public function __construct() {
 		$this->addType('created', 'int');
@@ -214,6 +216,8 @@ class Poll extends EntityWithUser implements JsonSerializable {
 				'orphanedVotes' => $this->getCurrentUserOrphanedVotes(),
 				'yesVotes' => $this->getCurrentUserYesVotes(),
 				'countVotes' => $this->getCurrentUserCountVotes(),
+				'yesVotesSub' => $this->getCurrentUserVotesYesSub(),
+				'countVotesSub' => $this->getCurrentUserVotesSub(),
 			],
 		];
 	}
--- a/lib/Db/PollMapper.php
+++ b/lib/Db/PollMapper.php
@@ -177,6 +177,8 @@ class PollMapper extends QBMapper {
 			->from($this->getTableName(), self::TABLE)
 			->groupBy(self::TABLE . '.id');
 
+		$qb->selectAlias($qb->createFunction('(' . $this->subQueryVotesCount(self::TABLE, $currentUserId)->getSQL() . ')'), 'current_user_votes_sub');
+		$qb->selectAlias($qb->createFunction('(' . $this->subQueryVotesCount(self::TABLE, $currentUserId, Vote::VOTE_YES)->getSQL() . ')'), 'current_user_votes_yes_sub');
 
 		$this->joinOptionsForMaxDate($qb, self::TABLE);
 		$this->joinCurrentUserVotes($qb, self::TABLE, $currentUserId);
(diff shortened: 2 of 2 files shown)


dartcafe:
The subquery seems not to work properly. When the query for one poll is executed the number of all votes is returns.

Return via API: 
```json
{
  "poll": {
    "currentUserStatus": {
      "yesVotesSub": 0,
      "countVotesSub": 7
    }
  }
}
```

However if I run the resulting SQL, the result is correct.

Resulting SQL query (by IQueryBuilder::getSQL()
```sql
SELECT `polls_polls`.*,
	(
		SELECT COUNT(`user_vote_sub`.`vote_answer`)
		FROM `*PREFIX*polls_votes` `user_vote_sub`
		WHERE (`user_vote_sub`.`poll_id` = `polls_polls`.`id`)
			AND (`user_vote_sub`.`user_id` = :dcValue1)
	) AS `current_user_votes_sub`,
	(
		SELECT COUNT(`user_vote_sub`.`vote_answer`)
		FROM `*PREFIX*polls_votes` `user_vote_sub`
		WHERE (`user_vote_sub`.`poll_id` = `polls_polls`.`id`)
			AND (`user_vote_sub`.`user_id` = :dcValue1)
			AND (`user_vote_sub`.`vote_answer` = :dcValue2)
	) AS `current_user_votes_yes_sub`,
	coalesce(MAX(options.timestamp), :dcValue1) AS max_date,
	coalesce(MIN(options.timestamp), :dcValue2) AS min_date,
	COUNT(`user_vote`.`vote_answer`) AS `current_user_votes`,
	coalesce(shares.type, :dcValue4) AS user_role,
	`shares`.`locked` AS `is_current_user_locked`
FROM `*PREFIX*polls_polls` `polls_polls`
	LEFT JOIN `*PREFIX*polls_options` `options` ON `polls_polls`.`id` = `options`.`poll_id`
	LEFT JOIN `*PREFIX*polls_votes` `user_vote` ON (`user_vote`.`poll_id` = `polls_polls`.`id`)
	AND (`user_vote`.`user_id` = :dcValue3)
	LEFT JOIN `*PREFIX*polls_share` `shares` ON (`polls_polls`.`id` = `shares`.`poll_id`)
	AND (`shares`.`user_id` = :dcValue5)
	AND (`shares`.`deleted` = :dcValue6)
WHERE `polls_polls`.`id` = :dcValue7
GROUP BY `polls_polls`.`id`,
	`shares`.`type`,
	`shares`.`locked`
```
results in 
```json
[
  {
    "current_user_votes_sub": "4",
    "current_user_votes_yes_sub": "2"
  }
]
```

dartcafe:
It seems, that the strings for the current user and the vote answer do not get applied inside the subquery
https://github.com/nextcloud/polls/blob/f7b663b5a5a39de54f60760affa71b6df71d6c73/lib/Db/PollMapper.php#L270
https://github.com/nextcloud/polls/blob/f7b663b5a5a39de54f60760affa71b6df71d6c73/lib/Db/PollMapper.php#L273

dartcafe:
OK. The IParam has to be defined outside of the subquery method.

dartcafe:
@come-nc Do you think it is a little bit overengeneered?
Are subqueries a good idea regarding performance?

come-nc:
Hello,

We looked into it and I think it would be best to avoid subqueries and limit ourselves to counts, sums and joins.
Do you think it would be possible to remove the orphan count information from the polls list? It’s the hardest to get without subquery or sidequery, and it does not look like something which should always be computed for all votes.

Couldn’t the orphaned votes simply be cleared at option deletion, or by a background job?

dartcafe:
> Couldn’t the orphaned votes simply be cleared at option deletion, or by a background job?

This is the problem. Deleted options are kept by intention, in case an option is deleted accitentially or is to be added again after deletion. User's votes should be kept in these cases. 

This is, why I think a subquery is better than running recursively over the result list. 

The other problem is that vote counts get multiplied by the avialable options before changing to the subquery. Sure, this could be solved in a join as well, but for me it works. 

Please take also a look into #3479. I tried to solve the frontend rendering problems by chunking the polls list and by reducing the polls in the navigation to a maximum of 6 entries.

dartcafe:
If the current situation suits your need, I am fine with releasing. I would prefer to concentrate on the master branch.

come-nc (approved):
Works and is better than previous state.
Having two separate subqueries for vote count and yes vote count seems suboptimal but let’s merge that for now it’s better anyway.
Maybe squash commits to help with backporting?

dartcafe:
> Having two separate subqueries for vote count and yes vote count seems suboptimal but let’s merge that for now it’s better anyway.

:-) I went the lazy day, and avoided figuring out how to selct two results from the subquery. Any hint for improvement is welcome.

dartcafe:
Ah damn, I forgot to squash. @come-nc

PR #3479, opened 2024-05-06
Enable chunk loading polls in list and limit polls in navigation.

dartcafe:
* render the polls list in chunks of 20 items
* continue chunk loading by scrolling to the end of the polls list
* Only load 6 items into the navigation sections
* Add counter to the navigation sections

Changed navigation |
-|
grafik|


susnux (approved):
Code looks okay. I personally would not implement this but reuse it from e.g. `@vueuse`.

dartcafe:
> Code looks okay. I personally would not implement this but reuse it from e.g. `@vueuse`.

I wanted to do that, but I would have been forced to use a quite old version; Last vue2 supporting version seems to be 6.x. Currently they are on version 10.x.

susnux:
> Last vue2 supporting version seems to be 6.x. Currently they are on version 10.x.

It supports Vue2 and Vue3 at the same time (they are using vue-demi), we use 10.x for some components on `@nextcloud/vue` :)

dartcafe:
Then I misinterpreted a comment in the docs. Anyways. ;-)

dartcafe:
@come-nc The performance win seems to be valid for backporting this to stable-6. What do you think? I would create a new beta right after it and then the release. 
@v1r0x Need you for a fast appstore release. The expectation is to have it before friday.

PR #3487, opened 2024-05-07
Switch from webpack to vite

dartcafe:
* Replacing webpack with vite and rollup
* remove Babel


dartcafe:
I did a new approach for switching from webpack to vite, but still staying on vue2 for the moment. 
Any idea for the right usage of the css implementation? 
This is the result
polls list |
-|
grafik |

ShGKme:
It seems like there is a bug in `vite-plugin-css-injected-by-js`. It only injects styles to the last entrypoint. In the current config — to the `dashboard`. If you put `main` to the end of entrypoints, it will work for Polls app main page (but not for other).

However, for apps, it's usually better to not inject CSS. This option is more useful for libraries.

You can disable `inlineCSS` and everything will work:
- Either without `cssCodeSplit: false` and then styles are separated into chunks and loaded from JS
- Or with `cssCodeSplit: false` but then you just need to manually load `css/polls-style.css` with aka 
  ```php
  Util::addStyle(AppConstants::APP_ID, 'polls-style');
  ```

Result will be around **16MB** JS + **0.2MB** CSS. In comparison, in `master` it is **60MB** JS.

Note, you have a large number of unneeded dynamic imports (dynamic components). It results in creating a large number of chunks and longer pages loading with a lot of network requests as well as slower rendering.

Dynamic import only makes sense for large + optional parts of the application. It doesn't make sense for super small components like SVG icon (`vue-material-design-icons`) or `PollItem`. It doesn't make sense for components that are almost guaranteed to be used on the page.

dartcafe:
> You can disable inlineCSS and everything will work:

OMG. I was so close all the time and such a stupid intention killed my efforts. 🤣 

> Note, you have a really large number of unneeded dynamic imports [...]

Thanks for the hint. I started to async load some components, which are rarely used, believing it would result in smaller and optimized chunks. Learned something.

Thanks for the hints, very valuable.

susnux:
> It seems like there is a bug in `vite-plugin-css-injected-by-js`... It only injects styles to the last entrypoint. In the current config — to the `dashboard`. If you put `main` to the end of entrypoints, it will work for the main page (but not for others).

It should work but needs `cssCodeSplit: true` so the vite metadata on where the styles where imported is populated.
And if you have multiple entry points the `relativeCSSInjection` option needs to be set to true: https://github.com/marco-prontera/vite-plugin-css-injected-by-js?tab=readme-ov-file#relativecssinjection-boolean

dartcafe:
> It should work but needs `cssCodeSplit: true` so the vite

The above setting currently works for me. I will try some options later, when I am at a stable state.

But I can't get the dev-tool add-on to work. 

Currently I get the following env variable: 
* import.meta.env.DEV=false 
* import.meta.env.PROD=true 
* import.meta.env.MODE development

Even, if I force the dev tools to `true`, they don't show up:
* Vue.config.devTools=true
* Vue.config.debug true

ShGKme:
> Any hint?

```
Vue.config.devtools = import.meta.env.MODE !== 'production'
```

src/js/main.js
```diff
@@ -27,23 +27,18 @@ import { sync } from 'vuex-router-sync'
...
-
 sync(store, router)
 
-Vue.config.debug = process.env.NODE_ENV !== 'production'
-Vue.config.devTools = process.env.NODE_ENV !== 'production'
-
+Vue.config.debug = true
+Vue.config.devTools = true
+console.debug('import.meta.env.DEV', import.meta.env.DEV, 'import.meta.env.PROD', import.meta.env.PROD, 'import.meta.env.MODE', import.meta.env.MODE, 'process.env.VUE_APP_DEBUG', process.env.VUE_APP)
+console.debug('Vue.config.devTools', Vue.config.devTools, 'Vue.config.debug', Vue.config.debug)
```
ShGKme, line 41: Some notes:
- `VUE_APP_` - is a prefix from Vue CLI. It is not available neither in Vite nor in any custom Webpack config other than Vue CLI service
- `MODE` and `DEV/PROD` have very different meaning:
  - `MODE` is about Vite building mode. By default, it is `production` for `vite build` and `development` for `vite`. But `--mode=X` overrides this. In other words, it is about how it is being bundled (using minification, debug and ect.). So this is what you want to use here
  - `DEV` and `PROD` only reflects `NODE_ENV` variable, with is always `development` for dev server (not used here) and `production` for building app `vite build` regardless of `--mode`.
  - See also: https://vitejs.dev/guide/env-and-mode.html#node-env-and-modes

ShGKme: Probably, we should manually set `NODE_ENV` to `development` when `--mode=development` in our vite config

dartcafe:
> It must be devtools, not devTools

OMG. They changed the casing? Thanks!" Works!

src/js/main.js
dartcafe: Thanks for the explanation.

dartcafe:
The app works now, so far so good, but now I ran in some weired issue, when usinf firefox (125 and 126). Chrome works without this issue. 

The vote page is empty, if I enter the poll directly with the URL i.e. `https://foo.com/apps/polls/vote/1` or make a reload, after I entered the page via link. 

I can see, that the fetched data is available inside the store in any case as usual. 

Entered via router link | After reload
-| -
grafik| grafik

Has anyone experienced this issue too? Or has an explanation?

ShGKme:
> Has anyone experienced this issue too? Or has an explanation?

I cannot reproduce... Seems like a cache issue.

dartcafe:
Correction. Now this happens on Chrome as well. Caching can be excluded:
I can reproduce this on different installations of Firefox (and now on Chrome) 

* Open Polls app `/apps/polls/list/relevant`
  * List appears, click on a poll and see the options and all the rest get loaded
* Clear the cache and all web storage
* Open Nextcloud
* Log in again
* Open polls `/apps/polls/list/relevant`
  * List appears
  * click a poll and see the options and all the rest get loaded `/apps/polls/vote/1`
* Reload page `/apps/polls/vote/1`
  * No store content is displayed, store if filled as before
* Enter poll URL in location bar `/apps/polls/vote/1`
  * No store content is displayed, store if filled as before
* Open poll via link in navigation `/apps/polls/vote/1`
  * List appears, click on a poll and see the options and all the rest get loaded

Additionally
* Open polls `/apps/polls/list/relevant`
  * List appears
* **Reload**
  * List appears
* click a poll `/apps/polls/vote/1`
    * No store content is displayed, store if filled as before

ShGKme:
@dartcafe using dev or pro build?

dartcafe:
I am using `dev`. But do you know what's funny? I removed everything and I had a very beautiful build. 

The only issue was, that the dashboard did not get any style, but it seems `inlineCSS: { relativeCSSInjection: true },` fixed that, too.

Seems I added some nonsense, which I did not understand. 😁

dartcafe:
@ShGKme If you mind, a short review would be great. And thanks for the support. 
@susnux Thanks, too.

ShGKme (approved):
Tested and seems to work fine

PR #3503, opened 2024-05-14
[stable-6] intoduce internal UserSession

come-nc:
Cherry-pick from master with conflicts

Backport of #3498

PR #3513, opened 2024-05-16
Add shares to poll schema

dartcafe:
## What has changed with this PR?

The PR addresses a few aspects. 
1. Performance. 
  Loading polls was rather slow with a big amount of polls, since there were a lot of database requests, mainly for permission checks and when dealing with the acl. 
  The new approach loads all necessary informations for the permission checks from the database via subqueries and joins. This way, permissions can be checked directly inside the poll. 
  As a result the old usage of the Acl has been changed and the new Acl carries not any poll informations, but the user information based on the `UserSession` and the app permissions. 
2. Move permission checks to the service layer, so the controllers are free from the permission checks.
3. Further steps to prepare creating polls without the necessity to be logged in. 
4. Preparation of the upcomming vue3 migration
  In the result the structure of the API has changed for a better segmentation. The most important change is, that all changable properties are collected inside the `configuration` node of the json structure. 

The following vue3 migration will be done in two steps:
    1. Migrating from vuex to pinia and typescript
    2. The final migration to vue3 with the options API including updating the stores
    3. Further rolling by-the-way migration of the options API to the composition API

@Bagoult FYI, since you are working with the API.

The structure layout is rather final, but may still get some updates in detail. Also the user objects will still receive some optimizations. I plan to get this API (v1.0) finalized. 

To dos: 

- [ ] Remove loading app settings from publicy accessing users
- [ ] Fix relevant listing
- [ ] Update user objects


dartcafe:
Unfortunately there is no dbal method to get an abtraction but the syntax is differetn. Therefore the platforms have to be identified and individual seletcs are necessary. 

@ChristophWurst Is there anyone who can verify this on sqlite, postgre and oracle? 

The setup should be a poll with some groupshares.
the excpected column should be `group_shares`= 'share1,share2'

dartcafe:
@come-nc @artonge The update was bigger than thought. I fear it will too error prone to backport that to the `stable-6` branch. 

@come-nc what do you think, would it be valuable to think about rebuilding the compatibility back until NC26/NC27 based on the master branch?

come-nc:
> @come-nc what do you think, would it be valuable to think about rebuilding the compatibility back until NC26/NC27 based on the master branch?

A colleague is planning to look into that. It may make sense given https://github.com/nextcloud/polls/pull/3503 is getting too complicated.
This is hoping the work is not more complicated than what was done to make stable-6 compatible with NC26.

dartcafe:
> A colleague is planning to look into that. It may make sense given #3503 is getting too complicated. This is hoping the work is not more complicated than what was done to make stable-6 compatible with NC26.

Yes. @artonge already told me. I think it will be done with replacing the attributes with annotations or just add them and by removing the few incompatibilities you already found. 

BTW: This PR will get some more updates, mainly for the js part.

dartcafe:
@artonge merging, thisd PR is huge enough. Any further fixes will be done in individual PRs.
New beta will be baked later this day

PR #3554, opened 2024-06-10
Remove service registration

### LINKED ISSUE #3551: Maintenance mode after update to Polls 7.1.0
### Announcement

Something is wrong with the NC updater I suspect, so this error leaves the system in maintenance mode. 

### You may find a log entry which says: 
```
TypeError: OCA\Polls\Db\UserMapper::__construct():
  Argument #2 ($userManager) must be of type OCP\IUserManager, OC\Session\Memory given, 
  called in lib/AppInfo/Application.php on line 158 and defined in lib/Db/UserMapper.php:60
```

Solution: run `occ upgrade` and disable the maintanance mode with `occ maintenance:moe --off`

See #3547 and #3526

dartcafe:
fix #3551

### DIFF
--- a/.github/ISSUE_TEMPLATE/ANNOUNCEMENT.yml
+++ b/.github/ISSUE_TEMPLATE/ANNOUNCEMENT.yml
@@ -5,4 +5,4 @@ labels: ["announcement"]
 body:
   - type: textarea
     attributes:
-      label: Annoncement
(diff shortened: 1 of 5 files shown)


dartcafe:
@v1r0x Are you able to push a new release to the appstore tonight? Version update will come later after all is checked in.

PR #3603, opened 2024-07-05
fix: strip illegal colon from sheet name

hamza221:
- Colon was missing from regex 
- error handling because it was failing silently


dartcafe (changes requested):
nice

dartcafe (approved):

Issue #3607, opened 2024-07-11
automatic reminder activated


lengnago:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### What went wrong, what did you observe?

The automatic reminder has been activated together with a new poll.
At the due date the following error is beeing logged:
[core] Fehler: Error while running background job OCA\Polls\Cron\NotificationCron (id: 19417, arguments: null)
	von ? von -- um 11.07.2024, 08:15:02
No reminder has been sent out to the parcipitating members. (The users have a stored mail address)
The same log entry is appears every 5 minutes

### What did you expect, how polls should behave instead?

A reminder mail has to be sent to the participating members

### What steps does it need to replay this bug?

The poll has been issued on July 06 2024
![Screen Shot 07-11-24 at 08 28 AM](https://github.com/nextcloud/polls/assets/149305544/abd54867-2833-48ee-914e-944f9bc2bbb9)

### Installation method

Installed/updated from the appstore (Apps section of your site)

### Installation type

Updated from a minor version within same major version (i.e. 4.0.0 to 4.1.1)

### Affected polls version

7.1.3

### Which browser did you use, when experiencing the bug?

- [X] Firefox
- [ ] Chrome
- [ ] Chromium/Chromium based (i.e. Edge)
- [ ] Safari
- [ ] Other/Don't know

### NC version

Nextcloud 29

### PHP engine version

PHP 8.2

### Database engine

MySQL

### Which user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other/Don't know

### Add your nextcloud server log here

```json
{
  "reqId": "Hop18pHCcLeKBqBilBEd",
  "level": 3,
  "time": "2024-07-11T06:35:01+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "core",
  "method": "",
  "url": "--",
  "message": "Error while running background job OCA\\Polls\\Cron\\NotificationCron (id: 19417, arguments: null)",
  "userAgent": "--",
  "version": "29.0.3.4",
  "exception": {
    "Exception": "OCP\\AppFramework\\Db\\DoesNotExistException",
[...]

PR #3608, opened 2024-07-11
fix: Catch exceptions thrown by NotificationMail constructor

come-nc:
Previously when constructor would throw, it would stop there and on the
 next run the notifications before the broken one would get resent.
 Exceptions from constructor are now caught and the process continues.

### DIFF
--- a/lib/Service/MailService.php
+++ b/lib/Service/MailService.php
@@ -156,15 +156,14 @@ class MailService {
 		}
 
 		foreach ($subscriptions as $subscription) {
-			$subscription->setNotifyLogs($this->logs);
-			$notication = new NotificationMail($subscription);
-
 			try {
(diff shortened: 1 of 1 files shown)


dartcafe (approved):

dartcafe:
Related to #3607?

AndyScherzinger:
@dartcafe Any change to get it merged and released as 7.1.x / 7.1.4? Thanks in advance

AndyScherzinger:
> Related to https://github.com/nextcloud/polls/issues/3607?

I guess in some way. The effect is that it retries this every 5 minutes and spams with mail notifications.

dartcafe:
I am just waiting for the merge and plan to release after it.

@come-nc Do you expect me to merge or do you prefer to merge it by yourself? I am fine with both approaches.

Issue #3630, opened 2024-08-05
Upgrade failed v7.1.4:  Undefined constant OCA\Polls\AppConstants::SESSION_KEY_CRON_JOB in UserSession.php:21


lucafemprocomuns:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### What went wrong, what did you observe?

Failed to upgrade Polls to version 7.1.4. using Nextcloud OCC tool.

### What did you expect, how polls should behave instead?

May be it is possible to fix adding:

use OCA\Polls\AppConstants;

on lib/UserSession.php:18:

### What steps does it need to replay this bug?

1. Upgrade from Nextcloud v27.1.10 to v27.1.11 with Polls v7.1.2 to v7.1.4 using php occ upgrade

### Installation method

Installed/updated from the appstore (Apps section of your site)

### Installation type

Updated from a minor version within same major version (i.e. 4.0.0 to 4.1.1)

### Affected polls version

7.1.2

### Which browser did you use, when experiencing the bug?

- [X] Firefox
- [ ] Chrome
- [ ] Chromium/Chromium based (i.e. Edge)
- [ ] Safari
- [ ] Other/Don't know

### NC version

Nextcloud 27

### PHP engine version

PHP 8.0

### Database engine

MySQL

### Which user-backends are you using?

- [ ] Default user-backend _(database)_
- [X] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other/Don't know

### Add your nextcloud server log here

```json
app_1  | Update app polls from App Store
app_1  | An unhandled exception has been thrown:
app_1  | Error: Undefined constant OCA\Polls\AppConstants::SESSION_KEY_CRON_JOB in /var/www/html/custom_apps/polls/lib/UserSession.php:21
app_1  | Stack trace:
app_1  | #0 /var/www/html/lib/private/AppFramework/Utility/SimpleContainer.php(84): ReflectionClass->newInstanceArgs(Array)
app_1  | #1 /var/www/html/lib/private/AppFramework/Utility/SimpleContainer.php(124): OC\AppFramework\Utility\SimpleContainer->buildClass(Object(ReflectionClass))
[...]

Issue #3604, opened 2024-07-06
Error while updating Polls


marienorbert:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### What went wrong, what did you observe?

While updating the Poll 7.1.2 to 7.1.3 in Nextcloud administration apps updating, Nextcloud is switching as normal to Maintanance mode and does not go on any further. Nextcloud stays blocked in maintanance mode.
All other apps are up to date.

### What did you expect, how polls should behave instead?

Normally there is no problem when updating.

### What steps does it need to replay this bug?

1. …/index.php/settings/apps/updates/polls
2. click on update
3. after about 5 to 10 seconds, Nextcloud is showing an error: "An error occurred during the request. Unable to continue. (translated from french: "Une erreur s'est produite pendant la demande. Impossible de poursuivre.")
4. Nextcloud stays blocked

### Installation method

Installed/updated from the appstore (Apps section of your site)

### Installation type

Updated from a minor version within same major version (i.e. 4.0.0 to 4.1.1)

### Affected polls version

7.1.2

### Which browser did you use, when experiencing the bug?

- [X] Firefox
- [ ] Chrome
- [ ] Chromium/Chromium based (i.e. Edge)
- [ ] Safari
- [ ] Other/Don't know

### NC version

Nextcloud 29

### Other Nextcloud version

Nextcloud Hub 8 (29.0.3)

### PHP engine version

PHP 8.2

### Database engine

MariaDB

### Database Engine version or other Database

Mysql 10.5.21

### Which user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other/Don't know

Mer0me:
You can run
`sudo -u www-data php occ maintenance:mode --off`
to regain access to Nextcloud.

But the bug must be resolved, it's not the normal behavior !

dartcafe:
Agree: https://github.com/nextcloud/server/issues/38158

PR #3631, opened 2024-08-05
Update UserSession.php

lucafemprocomuns:
Possible code for fix https://github.com/nextcloud/polls/issues/3630


dartcafe:
The reason is that class changes get not loaded properly while updating. See https://github.com/nextcloud/polls/issues/3604

A second update run fixes the issue.

PR #3648, opened 2024-08-14
fix : deleted user when email share registers

hamza221:
Reproduction:

1. Create a poll with a few options 
2. Share poll by mail 
3. open the share link received in the email 
4. Vote 
5. Register with the same email 

**Desired behaviour :** 
The new display name gets updated in the vote results 
**Actual behaviour :**
The vote is displayed as deleted user

### DIFF
--- a/src/js/components/Cards/VoteInfoCards.vue
+++ b/src/js/components/Cards/VoteInfoCards.vue
@@ -76,6 +76,7 @@ export default {
 			return (this.$route.name === 'publicVote'
 				&& ['public', 'email', 'contact'].includes(this.userRole)
 				&& !this.isPollClosed
+				&& !this.permissions.vote
 				&& !this.isLocked
 				&& !!this.pollId
 			)


AndyScherzinger (dismissed):

dartcafe:
I am struggeling to replay your Steps. 
What I did: 
1. - 3. 
After that I voted as the email owner. What do you mean with "register with the same email"?

hamza221:
After voting click Register 
image
And use the same address you received the invitation on

dartcafe:
Ahhhhhh. OMG. OK. You mean using the same share address after using it an initially vote. There is definetely a bug, but the bug is, that reusing the link leads to the register page and not directly to the poll. 

The share link is the direct ticket to enter the poll.

dartcafe:
The register method is the wrong address, since it is used to create a new share after registering and should only use the userId created there. Strange. I wonder why this leads to the registration page.

hamza221:
So if I understand correctly, the register button should only be shown for non logged-in public shares. 
It shouldn't be shown for mail shares since an ID already exists (the email address ) 
Correct?

hamza221:
may `currentUser` can be checked and the button should be rendered conditionally ? 
image
Should I update the Pr ?

dartcafe:
I have to reply my memory. 📝 

### First use case
A public link is used and the user has to register. He then gets a new share token and will be redirected to the poll with the fresh created share. 

### Second use case
An external user is invited via mail or via contact (like mail but with the contact's name as display name) from the contact app. 
In this case the user should be able to approve or change his email address and displayname first. I think this logic broke, when changing the workflow from first show the registration dialog to showing the poll without it and hiding the registration behind the register button. 

I think I forgot to respect the second use case whan changing the workflow.

dartcafe:
So changing the PR to quick fix that would be nice. Anyway, the userId is the email address, which is also not a good idea. It should be a gerneated id as well, since it is a personal public share. 

But I am thinking about making [next](https://github.com/nextcloud/polls/tree/next) the main branch, once the vue3 migration is stable and think about a more robust change over there.

AndyScherzinger (approved):

dartcafe (approved):

dartcafe:
It seems to do the job. But there is another problem. Changing the username or email address throws an error. 

Seems, I mixed up something somewhere. 🙈

sorbaugh:
Hello @dartcafe, hope it's ok that I merged and leave the newly arisen problem to be tackled in a further PR!

dartcafe:
Backport to next is not possible, since it got a complete overwork.

PR #3672, opened 2024-08-23
fix: broken email shares

hamza221:


lib/Migration/Version0721Date20240823135515.php
```diff
@@ -0,0 +1,64 @@
...
+			->fetchAll();
+
+		$userIds = array_map(fn ($row) => $row['user_id'], $userIds);
+
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select('*')
+			->from(Share::TABLE)
+			->where($qb->expr()->eq('type', $qb->createNamedParameter('external', IQueryBuilder::PARAM_STR)))
+			->andWhere($qb->expr()->notIn('user_id', $qb->createNamedParameter($userIds, IQueryBuilder::PARAM_STR_ARRAY)))
+			->andWhere($qb->expr()->in('email_address', $qb->createNamedParameter($userIds, IQueryBuilder::PARAM_STR_ARRAY)));
```
ChristophWurst, line 45: you might run into an IN with >1000 values and oracle doesn't like that. but I don't know a way to chunk this query tbh

lib/Migration/Version0721Date20240823135515.php
```diff
@@ -0,0 +1,64 @@
...
+
+		$affectedShares = $qb->executeQuery()->fetchAll();
```
ChristophWurst, line 47: closing the cursor missing again

dartcafe (changes requested):
I fear you will mix up conditions from different polls this way

hamza221:
> I fear you will mix up conditions from different polls this way

How so ? 
Can you recommend how to improve it? this doesn't have to be merged if you don't think it's necessary, we can just turn it into a query but we want to make sure that it's safe

dartcafe:
First of all, I do not fully understand the motivation behind this change. What do you want to achieve?

If I read the code right: All votes from all polls are taken and compared to all shares of all polls. 

The following combination can exist and would break your cleanup. 

votes: 
pollId | userId | ...
-|-|-
1 | bar | ...
2 | foo@bar.com | ...
15 | bar@foo.com | ...

shares
id | pollId | type |userId | emailAddress | ...
-|-|-|-|-|-
1 | 15 | external | bar | bar@foo.com | ...
2 | 16 | external | foo | foo@bar.com | ...

* Share 1 will not be altered because bar is existing from poll 1, but I believe is the actual target
* Share 2 will be altered, based on the existing vote from poll 4, but there is no indicator, that it should be

hamza221:
> What do you want to achieve?
fix broken shares from #3648 
> All votes from all polls are taken and compared to all shares of all polls. 

The logic is

1. Get all user ids in votes Table
2. check external shares (x)
3. if x.userId has 0 votes in all polls 
4. but x.emailaddress is the userId of at least one vote 
5. that means x is a broken share that is supposed to be a mail share not an external share 
6. Revert x to external and x.userId := x.emailaddress

> there is no indicator, that it should be 

The indicator is that no vote exists aka the logic 👆 

>Share 1 will not be altered because bar is existing from poll 1, but I believe is the actual target 

I'm not sure if Polls allows registering (or sending mail invites)  with an already registered email address
If it does I think it's another bug that should be fixed 
If it doesn't this can't happen

dartcafe:
I just listed possible combinations, which can exist in the wild. Maybe upon old bugs or intended but changed logic or deleted shares, or what ever. 

We don't know, if the problem is a wide spread problem besides this particular installation. The danger to change data unintentionally seems bigger to me than the value of this fix. 

It maybe cheaper to delete the users from the affected polls via the app.

dartcafe:
@hamza221 I am still nervous, that the fix could alter data unintentionally. But while I am thinking about the cause of this situation, I am very unpleased with the way the shares are used regarding the registration and email shares.

I think, I was too fast with the change then, when it was introduced and will overwork it completely in version 8.x. 

But what to do now with the situation and this PR. Is this still a thing in your support case? Maybe a manual cleaup would be more successful and safer than a cleanup with the risk of altering data on other installations out there. 

What do you think?

PR #3676, opened 2024-08-26
fix: disable vote option while loading

hamza221:
Without this you can run into this bug 
image
image

### DIFF
--- a/lib/Controller/VoteController.php
+++ b/lib/Controller/VoteController.php
@@ -53,7 +53,7 @@ class VoteController extends BaseController {
 			'vote' => $this->voteService->set($optionId, $setTo),
 			'poll' => $this->pollService->get($option->getPollId()),
 			'options' => $this->optionService->list($option->getPollId()),
-			]);
+		]);
 	}
 
 	/**
--- a/lib/Controller/VoteController.php
+++ b/lib/Controller/VoteController.php
@@ -8,6 +8,8 @@ declare(strict_types=1);
 
 namespace OCA\Polls\Controller;
 
+use OCA\Polls\Service\OptionService;
+use OCA\Polls\Service\PollService;
 use OCA\Polls\Service\VoteService;
 use OCP\AppFramework\Http\Attribute\NoAdminRequired;
 use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
@@ -21,7 +23,9 @@ class VoteController extends BaseController {
 	public function __construct(
 		string $appName,
 		IRequest $request,
-		private VoteService $voteService
+		private VoteService $voteService,
+		private PollService $pollService,
+		private OptionService $optionService,
 	) {
 		parent::__construct($appName, $request);
 	}
@@ -44,7 +48,12 @@ class VoteController extends BaseController {
 	#[NoAdminRequired]
 	#[NoCSRFRequired]
 	public function set(int $optionId, string $setTo): JSONResponse {
-		return $this->response(fn () => ['vote' => $this->voteService->set($optionId, $setTo)]);
+		$option = $this->optionService->get($optionId);
+		return $this->response(fn () => [
+			'vote' => $this->voteService->set($optionId, $setTo),
+			'poll' => $this->pollService->get($option->getPollId()),
+			'options' => $this->optionService->list($option->getPollId()),
+			]);
 	}
 
 	/**
--- a/lib/Service/OptionService.php
+++ b/lib/Service/OptionService.php
@@ -47,6 +47,9 @@ class OptionService {
 		$this->options = [];
 	}
 
+	public function get(int $optionId): Option {
+		return $this->optionMapper->find($optionId);
+	}
 	/**
 	 * Get all options of given poll
 	 *
--- a/lib/Service/VoteService.php
+++ b/lib/Service/VoteService.php
@@ -72,15 +72,20 @@ class VoteService {
 		$option = $this->optionMapper->find($optionId);
 		$poll = $this->pollMapper->find($option->getPollId());
 		$poll->request(Poll::PERMISSION_VOTE_EDIT);
-		
-		if ($setTo === Vote::VOTE_YES) {
-			$this->checkLimits($option);
-		}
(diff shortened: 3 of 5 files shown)


ChristophWurst (approved):
Looks good!

AndyScherzinger (approved):
LGTM, yet let's see what @dartcafe feedback will be.

dartcafe:
I can't reproduce it. How did you run into this situation?

hamza221:
> I can't reproduce it. How did you run into this situation?

It happens when you double click a vote option ( spamming also works)

dartcafe:
I klicked like a maniac, but the error did not raise, even, if I simulate a GPRS bandwidth.

But your solution has the side effect, that it lets the background flash unintenionally. 

https://github.com/user-attachments/assets/fb9993c4-3693-455b-8d3e-64fe2d7952aa

I see this issue as not necessary. It would be more valuable, if it would be applied to the next branch, because this branch will be a dead end. Next will replace master as future main branch.

dartcafe:
BTW: Any race condition should be solved by lodash/debounce and request canceling inside the api, so that only the last vote within the debounce timespan should be accepted. 

If this does not work for any reason, the fix should be applied there. Blocking the vote cell is not a good solution, since it is a bad UX.

hamza221:
>I klicked like a maniac, but the error did not raise, even if I simulate a GPRS bandwidth. 


https://github.com/user-attachments/assets/2f068f11-f61d-4888-9e7a-98c961af4dab

As you can see in the video the reproduction is pretty straight forward, You just have to double click a vote option 
You don't have to simulate a GPRS bandwidth. 

>since it is a bad UX. 

I'm sorry but I don't really understand you argument here. I think that the way it os right now is a bad UX (wrong warning message, and allowing a button press when you shouldn't ) 

I really think that a good UX in this case is 
1. Press Button 
2. Wait for action to fulfil 
3. allow action again 

If you think a loading indicator may help in this case we can add it

dartcafe:
First of all: I am still not able to provocate the 409.

But there is no reason to disable the current vote cell, with only one exception: The user already has **more** yes votes than allowed. In this case changing the vote from "yes" to "no" disables the current vote cell.

A race condition, caused by clicking while the request is still ongoing, should be captured by loadash/debounce or by cancelling previous request. In the result only the last vote is to be persisted.

This is, why I want to understand, under which circumstances the 409 occurs, because it should not.

dartcafe:
Buut. Luckily after "a million" clicks and gettint the exact click frequence, I now managed to get the 409 and see what happened. Wow, this is a very rare edge case with my setup.

Explanation:
The 409 is raised, because a yes vote is sent, but the server says, the vote limit is already reached, because this particular vote is already stored as "yes" and is therefore blocking the vote.

Currently this error is handeled by capturing it and reload the votes and the options to get a consistent state. But here we also have to reload the poll information, because the information about the remaining votes went there.

https://github.com/nextcloud/polls/blob/54334fe4e74f5a555ad1fdfa83b51c6b82a7da91/src/js/store/modules/votes.js#L107-L116

Now we can follow two strategies:
1. Avoid the 409 at the server side or
2. handle the error better at the client side, because now we have a misinformation about the remaining votes.
  The vote has the "yes" state, vote limit is set to "1 vote per user", but the card says "1 vote left", which is false. 

1. The vote is meant to get the state "yes", but it already has this state. This means we should skip the limit check and just commit the vote and return a response as if the vote has been set sucsessfully.

https://github.com/nextcloud/polls/blob/54334fe4e74f5a555ad1fdfa83b51c6b82a7da91/lib/Service/VoteService.php#L71-L85

2. Also reload the poll here additionally to the votes and options

https://github.com/nextcloud/polls/blob/54334fe4e74f5a555ad1fdfa83b51c6b82a7da91/src/js/store/modules/votes.js#L107-L116

dartcafe:
And thinking further about it, I believe both strategies should be applied. 
1. capture the situation, where the vote is meant to be set to the state it already has at the server side
2. load the poll configuration due to consistency reasons.

A further optimization would be to avoid two requests, by directly return the poll and options upon voting. 

I will directly add a commit, since I currently have the particular code opened.

dartcafe:
What I did: 
1. skip any action, if the stored vote already has the same state as it is intendet to be changed to 
2. revert your changes
3. Add poll loading if 409 should appear to secure consistency
4. Just optimization: Consolidate the three requests into one by returning poll and options list inside the vote request

dartcafe:
> I'm sorry but I don't really understand your argument here. I think that the way it is right now is a bad UX (wrong warning message, and allowing a button press when you shouldn't )

Since there wasn't any warning displayed to the user, I don't see any UX issue. What warning are you referring to? The bad UX was the misinformation about the left votes, which was wrong (you can see it in the last seconds of your video).

> I really think that a good UX in this case is
>     1. Press Button
>     2. Wait for action to fulfil
>     3. allow action again

And this is exatly what I intended to avoid and understand as a bad UX, since blocking prior votes is unnecessary. Only my last vote counts. 

For example in case of a laggy connection, it may be annoying, if I have to wait for the responses which are simply no more valid, to be able to set my decision of the vote. Therefore prior unfulfilled votes get discarded instead of waiting for the commits.

dartcafe:
@hamza221 I tested the changes quite some time: 
* I couldn't provocate the 429 anymore
* The vote cell stays clickable as intended (I think this is an information, that you were missing)
* Two requests get avoided after voting.

hamza221:
I can't reproduce anymore ✅

dartcafe:
OK. Great. Then I think, I can merge this...

dartcafe:
Just a recap: Understandable, why this happened. 

If the db state is yes and only the second vote ('no'-vote gets discarded) the conflict gets triggered, if the allowed number of remaining votes is 0. 

Then the limit check returns the exception, regardless if the vote does not change the number of used votes. 

Anyway strange I could not reproduce it from the start.

AndyScherzinger (approved):

AndyScherzinger:
Yes, has been quite a rabbithole to jump into 👍

PR #3677, opened 2024-08-26
fix: space in token causes 404 on logout

hamza221:
I noticed after loging out from a public share sometime a space character gets added to the token -> route which causes a 404  
I'm still not sure why it happens and it doesn't look deterministic, I'm still investigating 
image
Notice the extra space in the following screenshots
image
image

### DIFF
--- a/src/js/helpers/modules/cookieHelper.js
+++ b/src/js/helpers/modules/cookieHelper.js
@@ -56,7 +56,7 @@ const deleteCookieByValue = (searchValue) => {
 
 	if (cookieValue === searchValue) {
 		deleteCookie(cookieName)
-		return cookieName
+		return cookieName.trim()
 	}
 }
 


dartcafe:
Strange. Never saw that. How can a space get added to the cookie name, which is returned from the delete function. 

I think it better gets trimmed inside the function.

https://github.com/nextcloud/polls/blob/7e41c266f9e2820163afa20859e876f33650f3e9/src/js/helpers/modules/cookieHelper.js#L54C1-L61C2

dartcafe (changes requested):
Strange. Never saw that. How can a space get added to the cookie name, which is returned from the delete function. 

I think it better gets trimmed inside the function.

https://github.com/nextcloud/polls/blob/7e41c266f9e2820163afa20859e876f33650f3e9/src/js/helpers/modules/cookieHelper.js#L54C1-L61C2

hamza221:
> Strange. Never saw that. How can a space get added to the cookie name, which is returned from the delete function.
> 
> I think it better gets trimmed inside the function.
> 
> https://github.com/nextcloud/polls/blob/7e41c266f9e2820163afa20859e876f33650f3e9/src/js/helpers/modules/cookieHelper.js#L54C1-L61C2

I also couldn't figure out why it's happening

dartcafe (approved):

dartcafe:
🤷

PR #3689, opened 2024-09-03
feat: Keep unchanged indices and foreign keys instead of recreating

come-nc:
Hello,

As an alternative to switching to the standard way of handling schemas and indices, this PR avoid recreating already correct indices and foreign keys.

On foreign keys I’m less sure because I fail to see why they were recreated to start with, given that they are only removed on tables from `TableSchema::FK_CHILD_TABLES`, and I do not think the name of the parent table can ever change?

### DIFF
--- a/lib/Db/IndexManager.php
+++ b/lib/Db/IndexManager.php
@@ -122,11 +122,11 @@ class IndexManager {
 	 *
 	 * @return string[] logged messages
 	 */
-	public function removeAllForeignKeyConstraints(): array {
+	public function removeIncorrectForeignKeyConstraints(): array {
 		$messages = [];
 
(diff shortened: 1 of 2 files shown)


come-nc:
I should’ve duplicate the methods instead of renaming them, not sure how I missed that they were also used elsewhere -_-

dartcafe (commented):
Without checking this change by detail ATM. I had a call with @AndyScherzinger yesterday. I will start a new PR to be able to get a quick solution first, which can be pushedf to the appstore and with the latest fixes. 

I will link it after the first commit and decribe what my plan is. I will link it here and inside the original issue. 

Would be happy to get your review and maybe some tests are possible.

come-nc:
Closing this since indices are not dropped by the repair step anymore in master :+1:

Issue #3679, opened 2024-08-28
Drop and recreation of database indices on upgrade


ChristophWurst:
### ⚠️ This issue respects the following points: ⚠️

- [ ] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### What went wrong, what did you observe?

Upgrades are heavy on the database due to https://github.com/nextcloud/polls/blob/7e41c266f9e2820163afa20859e876f33650f3e9/appinfo/info.xml#L61-L68. Dropping an index is cheap. Recreating it is expensive.

### What did you expect, how polls should behave instead?

Use migration files or the *optional indices* APIs to add indices. Add them only once. Only drop them when they are not needed anymore.

### What steps does it need to replay this bug?

1. Install an older version of the app
2. Upgrade

### Installation method

Installed/updated from the appstore (Apps section of your site)

### Installation type

Updated from a minor version within same major version (i.e. 4.0.0 to 4.1.1)

### Affected polls version

v7.2.1

### Which browser did you use, when experiencing the bug?

- [ ] Firefox
- [X] Chrome
- [ ] Chromium/Chromium based (i.e. Edge)
- [ ] Safari
- [ ] Other/Don't know

### NC version

Nextcloud 27

### PHP engine version

PHP 8.1

### Database engine

MySQL

### Which user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other/Don't know

dartcafe:
Since the update may change the database shema, it is necessary to remove the indices, because of the constraints.

And since the update is done rarely I would say it is a acceptable cost.

sorbaugh:
> Since the update may change the database shema, it is necessary to remove the indices, because of the constraints.
> 
> And since the update is done rarely I would say it is a acceptable cost.

Hello @dartcafe ,

to give some context, a very large instance that uses the Polls App had a huge downtime when upgrading. Analyzing the problem we discovered that it is related to this indices situation.

The ideal way Nextcloud application manages the database schema is the following:

- For each version where the schema changes, create a Migration file to apply the changes to the schema. You can create tables, columns and indices in there.
- If you modify a table in a way that invalidates an index, then delete the index if present in the matching Migration file.

[...]

PR #3694, opened 2024-09-05
Avoid expensive index recreation

dartcafe:
**This is a hotfix** to address a particular problem with heavy and expensive index recreation on large databases.

relates #3679 but not final fix
backlink to #3689 for transparancy

Problem description: 
At least one known instance with a large amount of polls and votes, etc. suffers from the recreation of indices, because it slows down the database after every update. Oviously the database discards existing execution plans and in the result, the database slows down until the execution plans are optimized again. 

This is the problem how I understood it. Correction are welcome. 

A quick analysis of the current code reveals, that it is safe to keep the index creation, since existing indices seem to stay untouched. This affects unique and other indices (check for existing indices already exists) and foreign key constrains as well. So it seems, that the removal of the pre-migration repair step alone will prevent an index recreation.

Currently this is safe for most of the installations, since the last table change, where a prior index removal was neccessary, is long time ago. I will check this later in detail, how long ago. 

If neccesary I plan to add outdated migrations to the latest existing migration, to ensure that this is only applies for updates from Versions before Polls 6.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 -->
 # Changelog
 ##
+## [7.2.2] - 2024-09-06
+### Fix
+ -  fix watcher in situations it may fail on pollId 0
+ -  fix failing routes on tokens with trailing spaces
+ -  Removed index removal from the pre-migration repair steps
+
 ## [7.2.1] - 2024-08-22
 ### Fix
  -  Fix deleted user when email share registers
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -58,7 +58,6 @@
   <repair-steps>
     <pre-migration>
       <step>OCA\Polls\Migration\RepairSteps\RemoveObsoleteMigrations</step>
-      <step>OCA\Polls\Migration\RepairSteps\RemoveIndices</step>
     </pre-migration>
     <post-migration>
       <step>OCA\Polls\Migration\RepairSteps\DropOrphanedTables</step>
--- a/lib/Command/Db/Rebuild.php
+++ b/lib/Command/Db/Rebuild.php
@@ -23,6 +23,12 @@ class Rebuild extends Command {
 	protected array $operationHints = [
 		'All polls tables will get checked against the current schema.',
 		'NO data migration will be executed, so make sure you have a backup of your database.',
+		'',
+		'*****************************',
+		'**    Please understand    **',
+		'*****************************',
+		'The process will also recreate all indices and foreign key constraints.',
+		'This can lead to a database performance impact on the app after the recreation is done.',
 	];
 
 	public function __construct(
--- a/lib/Db/IndexManager.php
+++ b/lib/Db/IndexManager.php
@@ -73,7 +73,7 @@ class IndexManager {
 	/**
 	 * add an on delete fk contraint
 	 *
-	 * @param string $parentTableName name of reffered table
+	 * @param string $parentTableName name of referred table
 	 * @param string $childTableName name of referring table
 	 * @return string log message
 	 */
--- a/lib/Service/PreferencesService.php
+++ b/lib/Service/PreferencesService.php
@@ -30,7 +30,7 @@ class PreferencesService {
 	public function load(): void {
 		try {
 			$this->preferences = $this->preferencesMapper->find($this->userSession->getCurrentUserId());
-		} catch	(Exception $e) {
+		} catch (Exception $e) {
 			$this->preferences = new Preferences;
 		}
 	}
--- a/lib/Service/WatchService.php
+++ b/lib/Service/WatchService.php
@@ -37,7 +37,7 @@ class WatchService {
 	 */
 	public function watchUpdates(int $pollId, ?int $offset = null): array {
 		
-		if($pollId !== 0) {
+		if ($pollId !== 0) {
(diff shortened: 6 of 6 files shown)


artonge (dismissed):
Shall we also delete `lib/Migration/RepairSteps/RemoveIndices.php`?

hamza221 (approved):

hamza221:
Let's wait for @Altahrim and @ArtificialOwl to have a look before merging  🙏

ArtificialOwl (dismissed):

ArtificialOwl:
i am down for any cleaning of pre/post migration steps !

Altahrim (dismissed):

dartcafe:
> i am down for any cleaning of pre/post migration steps !

I will tidy theese things up later. But I found another problems, the foreign key constraints don't get created, if removed once. Have to get into that.

AndyScherzinger:
I suppose the same conceptual logic applies for the FKs? Don't recreate if already there (smart manager bla, sicne I personally lack Nc apps dev knowhow beyond the mobile apps...)

dartcafe:
> I suppose the same conceptual logic applies for the FKs? Don't recreate if already there (smart manager bla, sicne I personally lack Nc apps dev knowhow beyond the mobile apps...)

The problem is, I have no idea how to check, if the FK already exists or not. 

I plan to get away from it, but the logic of removing dependent rows from other tables has to be moved into the app in this case. At least into the janitor cron.

hamza221:
I'm sorry I'm just trying to further understand, what the problem is so maybe I can try help: 

- The foreign keys and indices used to both be dropped the `RemoveIndices`the step which is not used anymore so why is it needed to recreate them. 
> The problem is, I have no idea how to check, if the FK already exists or not.
- You mean before creating them here https://github.com/nextcloud/polls/blob/78b97c28309ae84b6f1b23a1b6ad3cd09d20fc56/lib/Db/IndexManager.php#L80-L88 ?  If so can't [listTableForeignKeys()](https://www.doctrine-project.org/projects/doctrine-dbal/en/4.1/reference/schema-manager.html#listtableforeignkeys) be used ?

dartcafe:
> * The foreign keys and indices used to both be dropped the `RemoveIndices`the step which is not used anymore so why is it needed to recreate them.

They don't need to be recreated. Just in case they are not there. 

> ? If so can't [listTableForeignKeys()](https://www.doctrine-project.org/projects/doctrine-dbal/en/4.1/reference/schema-manager.html#listtableforeignkeys) be used ?

🤣 This is what I missed. Thanks. 

But the situation is, if the FK is requested to be created and is already present the commend gets ignored and the FK index stays untouched. 


But I tricked myself or I am too stupid to recognize, that the FK indices get created. False alarm.

dartcafe:
Summary:
* removing indices has been removed from the steps
* index creation is still present as before, but should not affect existing indices

@AndyScherzinger I believe that should work for your use case.

dartcafe:
I now insprected the changes until 16 months ago. Since then no change to the database was made, which could be a problem to the indices (and makes it neccessary to rebuild them). 

If some one with an old version (prior v5) runs into an update issue, I will take care of it.

dartcafe:
> and a final question, Do you know why they can't be created? (a log or an error message)

@hamza221 I just confused myself, beacuase I was checking the migrations with a remote server and check the database with phpMyAdmin. And I expected the fk_* indices to appear together with the other indices. 

Until I realized, the poll_id was linked to the polls table and remembered that the KF indices are visible inside the relation view. 

Layer 8 problem. 😁 

I think it is time to install a nextcloud test and dev instance locally.

AndyScherzinger:
@dartcafe sounds good to me (7.2.2 release and publishing) folks are already out but I see to have them test an update 7.1.4 to 7.2.2 on Monday then as a final step before suggesting the upgrade.

@hamza221 could you do a final test with sql logging on db to check no index nor fk gets (re) created? Thanks in advance.

AndyScherzinger (approved):

AndyScherzinger (approved):
Looks good to me 👍

Thanks a lot again for taking care of this matter 🎉

AndyScherzinger:
@dartcafe just to be safe here since I can't tell if you left out the last step (publishing to the store) on purpose. So for automated publishing you'll need to push the tag 7.2.2 of this repository to https://github.com/nextcloud-releases/polls where you'd also create a release based on it.

The automation should then take care of the app store publishing based on that tag/release.

I just wanted to check since I am not sure if you wanted to wait until folk had another testing round on Monday.

dartcafe:
@AndyScherzinger I an just not sure how to do that. Is it really meant by doing the steps manually? 
* Create Tag v7.2.2
* add release manually to the releases by using the dialog?

dartcafe:
Nope. This did not work. 

What exatctly is meant by "push the tag 7.2.2 of this repository"? Nerver pushed a tag from one repo to another.

dartcafe:
Yeah ok. it was `git push https://github.com/nextcloud-releases/polls v7.2.2`

AndyScherzinger:
Would have had to. Do it myself to tell you the git command since I use GUI git clients only 🙂

AndyScherzinger:
Also just checked the app store and 7.2.2 as well as 8.0.0-alpha4 are showing 👍

PR #3714, opened 2024-09-19
fix: show edit inputs only for external shares

hamza221:
Steps to reproduce:

1. Create a poll with some options
2. Go to share tab and share vial mail
3. Open the poll with that share link
4. open the settings 
5. seem them turn green
6. Hit the fields' save arrow at the right
7. `Displayname can only be set for external shares.`

### DIFF
--- a/lib/Db/Share.php
+++ b/lib/Db/Share.php
@@ -69,12 +69,16 @@ class Share extends EntityWithUser implements JsonSerializable {
 	public const TYPE_CONTACTGROUP = 'contactGroup';
 
 
+	public const CONVERATABLE_PUBLIC_SHARES = [
+		self::TYPE_EMAIL,
+		self::TYPE_CONTACT,
+	];
+
 	// Share types, that are allowed for public access (without login)
 	public const SHARE_PUBLIC_ACCESS_ALLOWED = [
 		self::TYPE_PUBLIC,
-		self::TYPE_CONTACT,
-		self::TYPE_EMAIL,
 		self::TYPE_EXTERNAL,
+		...self::CONVERATABLE_PUBLIC_SHARES,
 	];
 
 	// Share types, that are allowed for authenticated access (with login)
--- a/lib/Service/ShareService.php
+++ b/lib/Service/ShareService.php
@@ -36,9 +36,15 @@ use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\DB\Exception;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\Security\ISecureRandom;
+use phpDocumentor\Reflection\Types\This;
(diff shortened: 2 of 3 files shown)


dartcafe:
This will cover only 'external shares'. 
Should be also allowed for email shares. 

Maybe computed properties (isAllowedChangeDisplayName, isAllowedChangeEmailAddress) are a better style. Not sure if.

dartcafe (changes requested):
This will cover only 'external shares'. 
Should be also allowed for email shares AFAIR, too. 

Maybe computed properties (isAllowedChangeDisplayName, isAllowedChangeEmailAddress) are a better style. Not sure if.

hamza221:
@dartcafe  the backend is throwing an error if share type is not `Share::TYPE_EXTERNAL`
https://github.com/nextcloud/polls/blob/master/lib/Service/ShareService.php#L196-L209

That's how I came up with the conclusion that it should only be visible for external shares, I'm not aware of any limitation on why email shares are allowed
If you think we should also allow it for `Share::TYPE_EMAIL` I can revert the changes and make it into a backend fix

dartcafe:
Ah. OK. now I understand your problem. Maybe it is a good idea to first file a bug issue. 

So, if I understood you right, the problem is the rejection of changing the name of an email share. This should be possible, which seems to identify a bug at the backend side. 

I have to check this, since AFAIR an email share should be converted to an external share after first call via this share. Similar to creating a new share, when entering the poll via public share. 

The sharing system seems still to have some flaws, since some changes a few months ago.

AndyScherzinger:
> So, if I understood you right, the problem is the rejection of changing the name of an email share. This should be possible, which seems to identify a bug at the backend side.

Yes. So basically when you share a poll via an email address and open the poll with that link. When you try to change the mail address or the name you will be always facing an http 403 or 409

2024-09-19 10_19_55-NVIDIA GeForce Overlay

In the screenshot I changes the mail address and the name from the original values from the share link. they turn green while typing, but hitting the send arrow icon you face the errors seen in the browser console

dartcafe:
Yes, I could reproduce the bug. But I still think, it should be fixed on backend side, since the rejections is the error. Email shares should also be allowed to change name and email address.

I fear the cause is the missing registration (it was removed for these shares by hiding the registration dialog out of convenience) in case of email shares. 

And I assume, contact shares are affected as well.

dartcafe (approved):
Both changes together do their work.

Accidentally early merge of the follow up PR.

dartcafe:
I've tested the changes: 
* E-Mail and contact shares get converted to an external share at first usage
* These users are again able to change their names and mail addresses 

I will merge this, since I want the update to get out quickly.

PR #3716, opened 2024-09-21
Convert email and contact shares to external shares on first usage

dartcafe:
Explanation of this fix, which should be appied to #3714 first.

The cause of this behavior is #3125. This change removed the initial registration dialog after entering a poll via share token. The dialog was meant to allow participants 
* to register while using a public share token and create a personal share and 
* to check user parameters (mail and displayname) using an email or contact share. 

The second use case was eliminated with the issue causing PR. The result was, that the share type stays the initial type. 

This PR now converts theese shares into external sharesm by changing the type and replacing the userId by a generated one, to make sure, sensitive informations get removed (like mail address in email shares).

### DIFF
--- a/lib/Db/Share.php
+++ b/lib/Db/Share.php
@@ -69,12 +69,16 @@ class Share extends EntityWithUser implements JsonSerializable {
 	public const TYPE_CONTACTGROUP = 'contactGroup';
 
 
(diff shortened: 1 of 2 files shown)


dartcafe (commented):
/backport to next

dartcafe:
damn. Did not wanted to merge it now. sry

PR #3777, opened 2024-11-14
Fix user search by email

come-nc:
Searching for users to share with by email was broken.
This is due to 2 problems:

1. The subname key for users had the wrong casing `subName` instead of `subname`
2. The frontend filtering was active despite not being needed as backend already returns a filtered list

Before  | After
------------- | -------------
Copie d'écran_20241114_175310  | Copie d'écran_20241114_175142

@dartcafe Is it safe to rename subName to subname or is used elsewhere? The only instance I could find is https://github.com/nextcloud/polls/blob/master/src/js/store/modules/acl.js#L16 and it looks pretty safe to rename if needed.

### DIFF
--- a/lib/Model/UserBase.php
+++ b/lib/Model/UserBase.php
@@ -308,14 +308,14 @@ class UserBase implements JsonSerializable {
 	 *
 	 * @return (bool|string|string[])[]
 	 *
-	 * @psalm-return array{userId: string, displayName: string, emailAddress: string, subName: string, subtitle: string, isNoUser: bool, desc: string, type: string, id: string, user: string, organisation: string, languageCode: string, localeCode: string, timeZone: string, icon: string, categories: array<string>}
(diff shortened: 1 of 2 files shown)


dartcafe:
Renaming the subname seems to be no problem. 
As far as I remember, removing the filter removed the highlighting as well, since the match may also occur i.e. with the company name of a contact. But this seems to be no more the case.

dartcafe (approved):

dartcafe:
Just for reference: https://github.com/nextcloud-libraries/nextcloud-vue/issues/6219

come-nc:
@dartcafe Can this be merged? Should I ask the bot to "backport" to `next`?

dartcafe:
Sure, I approved. :wink:

dartcafe:
@come-nc Ah, no, backport not necessary. I did the backport already manually, just forgot to link that.

come-nc:
@dartcafe Would you consider releasing 7.2.5 so that this fix makes it to the appstore?

dartcafe:
@come-nc I released 7.2.5 yesterday.

PR #3878, opened 2025-02-20
fix: Remove dropdown options limit from sharing search

### LINKED ISSUE #3864: Share dropdown does not show all matches

- [x] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [x] This issue is **not** already reported on Github _(I've searched it)_.

When searching for group "abc" but there are other groups, e.g. abc1, abc2, abc3, etc., that also match before it you can't select group "abc" as only a limited number of matches are searched for

All matches should be searched for and not stop at an arbitrary limit

1. Create 40 groups abc, abc1, abc2, abc3, etc.
2. Create a poll
3. Search for "abc" in share dropdown

### Installation method

Installed/updated using occ

### Installation type

First time installation

### Affected polls version

master

- [ ] Firefox
- [ ] Chrome
- [ ] Chromium/Chromium based (i.e. Edge)
- [ ] Safari
- [ ] Other/Don't know

### Other browser

_No response_

### Add your browser log here

```json

```

_No response_

### NC version

Other/Don't know

### Other Nextcloud version

master

### PHP engine version

PHP 8.2

### Other PHP version

_No response_

### Database engine

MySQL

_No response_

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other/Don't know

```json

```

_No response_

### Configuration report

```json

```

### List of activated Apps

```shell

```

### Nextcloud Signing status

```shell

```

### Additional Information

_No response_

Pytal:
Fix #3864

### DIFF
--- a/src/js/components/User/UserSearch.vue
+++ b/src/js/components/User/UserSearch.vue
@@ -11,7 +11,6 @@
 		:user-select="true"
 		:filterable="false"
 		:tag-width="80"
-		:limit="30"
 		:loading="isLoading"
 		:searchable="true"
 		:placeholder="placeholder"


dartcafe (approved):
As we currently limit the amount of results to 100 by the backend we can remove the limit. I just wonder, what happens, if we have 1000 or more results. I guess some lazy loading mechanism has to be applied then. 

Lets try until then.

dartcafe:
I'll merge it, since I think, you don't want to add anything more.

Pytal:
Thanks for merging @dartcafe 🚀

> As we currently limit the amount of results to 100 by the backend we can remove the limit. I just wonder, what happens, if we have 1000 or more results. I guess some lazy loading mechanism has to be applied then.

At 1000 results it would probably make more sense to have more distinct names so they can be searched for directly

PR #3881, opened 2025-02-25
Fix excessive vote column width

asdil12:
Only make the columns as wide as needed.
Allow them to grow if there is unused space left.
Do not allow them to grow and then show a horizontal scrollbar.

Before:
image

After:
image

Horizontal scrolling still works if there is really not enough space:
image

### DIFF
--- a/src/js/components/VoteTable/VoteTable.vue
+++ b/src/js/components/VoteTable/VoteTable.vue
@@ -139,7 +139,7 @@ export default {
 	.vote-column {
 		order: 2;
 		display: flex;
-		flex: 1 0 11em;
+		flex: 1 0 auto;
 		flex-direction: column;
 		align-items: stretch;
 		max-width: 19em;


dartcafe (changes requested):
Please add tests containing more variations:
* date polls with date and time ranges
* a mixture of them
* Text polls

asdil12:
Date polls:

Old:


New:

asdil12:
Not sure what you mean with _a mixture of them_ but it doesn't seem possible to have both text and date options in a single poll.

dartcafe:
> Not sure what you mean with a mixture of them.

I meant what is displayed with the last screenshot. 

But your screenshots display the reason, why the width is set. With different widths of the header content also the columns will have different widths. 

This should be avoided from an UX perspective, since the different width may signal different weights and importance of the options. 

So the goal was (and should be kept) to have identical widths of the columns. I have to admit that the current solution was a shortcut to achieve that. But with an `auto` setting, we still have the initial situation.

If you have a solution, to ensure, that the width is set even over all columns by saving space, we should try it.

pheerai:
In general, the requirement for "All option columns should have the same width" somewhat indicates to me that a `grid` would be a better fit compared to `flex`.

The following seems to work fine for me:

 * In `.vote-table__users`: `flex: 0 fit-content`
 * In `.vote-table__votes`: `display: grid` instead of `flex` and `grid-template-columns: repeat(999, max-content)`

Are there any edge-cases that break through the `max-content` here that I'm not seeing?

Of course, the hardcoded `999` needs to be replaced with the actual number of options, but I'm not well-versed enough with Vue to know whether that's something one would do (and, if so, how exactly).

dartcafe:
@pheerai Yeah. From time to time, I was thinking about changing it to a grid view, but had not the mood or time to get into it, since I did not used grid views so far.

dartcafe:
I will discard this PR.
Reasons: 
* v7 will slowly become legacy, since the v8 is nearly about to leave beta state
* v8's code base is not compatible to v7's, so any changes will have a short life
* columns with different widths are to be avoided. Actually having whitespace is a good practice and should be preferred over packed information

PR #3893, opened 2025-03-03
enh: respect sharing restriction in sharing admin config

hamza221:
image

Next step: respect the rest of sharing enumerations

### DIFF
--- a/lib/Service/SystemService.php
+++ b/lib/Service/SystemService.php
@@ -23,6 +23,9 @@ use OCA\Polls\Model\User\Contact;
 use OCA\Polls\Model\User\Email;
 use OCA\Polls\Model\User\User;
 use OCP\Collaboration\Collaborators\ISearch;
+use OCP\IConfig;
+use OCP\IGroupManager;
+use OCP\IUserSession;
 use OCP\L10N\IFactory;
 use OCP\Share\IShare;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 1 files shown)


lib/Service/SystemService.php
```diff
@@ -243,4 +253,22 @@ public function validatePublicUsername(string $userName, Share $share): string {
...
+	private function isUserAllowedToShare(): bool {
+		$excludedGroups = $this->config->getAppValue('core', 'shareapi_exclude_groups_list', '');
+		if ($excludedGroups === '') {
+			return true;
+		}
+		$excludedGroups = json_decode($excludedGroups, true);
+
+		$user = $this->userSession->getUser();
+		if ($user) {
+			if ($excludedGroups) {
```
icewind1991, line 266: nitpick: these two if statements can be merged

dartcafe:
The backport reques is just a reminder, since the backport will fail. I will add it manually to next

sorbaugh:
If I understand correctly, this PR is sufficient for Polls 7.x, but that for Polls 8.x a better solution is strongly recommended. Can we merge this and replace it with a better solution later?

Ideally we'd need to have this (or a better) solution for march 10th. Of course, being respectful of your time, if that weren't possible we can use this as a patch for now (and would need to keep track of it) 👍

dartcafe (changes requested):
Prefering #3894

PR #3894, opened 2025-03-06
Apply core share group restrictions

dartcafe:
Alternative implementation of #3893

For reviewers and testers, please check the checkboxes, if the functionality can be approved. I could confirm the functionality so far. After approval, I will merge and build version 7.4.

Summary: 
- [ ] Deny creation of shares if user is member of a group which is excluded from created shares by `limit sharing based on groups` set to `exclude some groups from sharing` 
grafik
- [ ] Allow creation of shares only for users which are members of a group which is allowed to create shares by `limit sharing based on groups` set to `limit sharing to some groups` 
grafik
- [ ] Allow sharing to everyone by `limit sharing based on groups` set to `Allow sharing for everyone (default)` 
grafik

Additionally 
- [ ] Deny all external shares based on `Allow users to share via link and emails` set to `off` 
grafik
- [ ] Deny external shares for excluded groups based on `Allow users to share via link and emails` set to `on` and `Exclude groups from creating link shares` 
grafik

External shares in this context are
* public links
* email shares
* contacts and contact groups

Circles (teams) will still be allowed, but external users are skipped, if the Circle (team) gets resolved. 

Note: Polls still uses the term Circle

### DIFF
--- a/lib/Model/Settings/AppSettings.php
+++ b/lib/Model/Settings/AppSettings.php
@@ -145,23 +145,35 @@ class AppSettings implements JsonSerializable {
 		$groupExceptionMode = $this->getShareGroupExceptionMode();
 
 		if ($groupExceptionMode === 'off') {
-			// no group exceptions
+			// no group exceptions are set, allow share creation
 			return true;
+
+			// TODO: get clarification for the usage of the excluded groups
+			// The following code is not reachable by intention, because it is based on a misunderstanding.
+			// just kept for the moment as reminder
+			//
+			// Additionally check, if user is in excluded groups 'Exclude groups from creating link shares'
+			$excludedGroups = $this->getShareExcludedGroups();
+			if ($excludedGroups) {
+				// if user is in exception group, allow share creation
+				return !$this->userSession->getUser()->getIsInGroupArray($excludedGroups);
+			}
 		}
 
 		// get group exceptions
-		$excludedGroups = $this->getShareExceptionGroups();
+		$exceptionGroups = $this->getShareExceptionGroups();
 
 		if ($groupExceptionMode === 'allowGroup') {
-			// exception mode is 'limit sharing to some groups'
+			// exception mode is 'Limit sharing to some groups'
 			// if user is in exception group, allow share creation
-			return $this->userSession->getUser()->getIsInGroupArray($excludedGroups);
+			return $this->userSession->getUser()->getIsInGroupArray($exceptionGroups);
+		} elseif ($groupExceptionMode === 'denyGroup') {
+			// exception mode is 'Exclude some Groups from sharing'
+			// if user is in exception group, deny share creation
+			return !$this->userSession->getUser()->getIsInGroupArray($exceptionGroups);
 		}
 
-		// exception mode is 'Exclude some Groups from sharing'
-		// if user is in exception group, deny share creation
-		return $this->userSession->getUser()->getIsInGroupArray($excludedGroups);
-
+		return true;
 	}
 
 	/**
@@ -169,7 +181,16 @@ class AppSettings implements JsonSerializable {
 	 * @return array
 	 */
 	public function getShareExceptionGroups(): array {
-		$excludedGroups = $this->config->getAppValue('core', 'shareapi_exclude_groups_list', '');
+		$exceptionGroups = $this->config->getAppValue('core', 'shareapi_exclude_groups_list', '');
+		return json_decode($exceptionGroups, true) ?? [];
+	}
+
+	/**
+	 * Get share excluded groups
+	 * @return array
+	 */
+	public function getShareExcludedGroups(): array {
+		$excludedGroups = $this->config->getAppValue('core', 'shareapi_allow_links_exclude_groups', '');
 		return json_decode($excludedGroups, true) ?? [];
 	}
 
@@ -177,13 +198,13 @@ class AppSettings implements JsonSerializable {
 	 * Get share group exception mode
 	 * @return string
 	 * @psalm-return 'denyGroup'|'allowGroup'|'off'
-	 * Take value from the core setting 'shareapi_exclude_Mode' and translate
+	 * Take value from the core setting 'shareapi_exclude_groups' and translate
 	 * 'yes' => 'denyGroup' ('Exclude some groups from sharing')
 	 * 'allow' => 'allowGroup' (Limit sharing to some groups)
 	 * default => 'off' (Allow sharing for everyone) or setting absent
 	 */
 	public function getShareGroupExceptionMode(): string {
-		$excludedMode = $this->config->getAppValue('core', 'shareapi_exclude_Mode', '');
+		$excludedMode = $this->config->getAppValue('core', 'shareapi_exclude_groups', '');
 		return match ($excludedMode) {
 			'yes' => 'denyGroup',
 			'allow' => 'allowGroup',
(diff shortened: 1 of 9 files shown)


dartcafe:
Follow up to #3893 

If I understand the settings right, it should work the following way: 

* `Allow sharing for everyone` -> Everybody is allowed to create shares as a poll owner or delegated poll admin
* `Limit sharing to some groups` -> Only listed groups are allowed to create shares as a poll owner or delegated poll admin
* `Exclude some groups from sharing` -> Listed groups are denied to create shares as a poll owner or delegated poll admin

Assuming a right understanding, I moved the check to the appSettings, by getting the mode and apply the listed groups accordingly

Please be so kind and approve this assumption shortly.

dartcafe:
I need some additional explanation: 
How does this setting play with `Exclude groups from creating link shares`?

image

hamza221:
Setting `exclude some groups from sharing`, users in the groups are still able to search and share with other users

lib/Model/Settings/AppSettings.php
```diff
@@ -132,6 +132,65 @@ public function getComboAllowed(): bool {
...
+	 * Get share group exception mode
+	 * @return string
+	 * @psalm-return 'denyGroup'|'allowGroup'|'off'
+	 * Take value from the core setting 'shareapi_exclude_Mode' and translate
+	 * 'yes' => 'denyGroup' ('Exclude some groups from sharing')
+	 * 'allow' => 'allowGroup' (Limit sharing to some groups)
+	 * default => 'off' (Allow sharing for everyone) or setting absent
+	 */
+	public function getShareGroupExceptionMode(): string {
+		$excludedMode = $this->config->getAppValue('core', 'shareapi_exclude_Mode', '');
```
hamza221, line 186: Where is this config coming from, cant seem to find it in the server repo

lib/Model/Settings/AppSettings.php
```diff
@@ -132,6 +132,65 @@ public function getComboAllowed(): bool {
...
+	public function getShareCreateAllowed(): bool {
+		if (!$this->userSession->getIsLoggedIn()) {
+			// only logged in users can create shares
+			return false;
+		}
+
+		// first check group exception mode
+		$groupExceptionMode = $this->getShareGroupExceptionMode();
+
+		if ($groupExceptionMode === 'off') {
```
hamza221, line 147: While testing with `shareapi_exclude_groups_list` containing values, the function returns early here, without checking for excluded groups and allows users in the group to find other users  and create shares

hamza221 (changes requested):

lib/Model/Settings/AppSettings.php
dartcafe: I guess some hallucination :rofl:. Should be `shareapi_exclude_groups`

artonge:
> I need some additional explanation: How does this setting play with `Exclude groups from creating link shares`?

Which setting? `Limit sharing based on group`?

1. `Limit sharing based on group`: Limit internal and external sharing
2. `Exclude groups from creating link shares`: Prevent members of the listed group to create link shares.

N°1 takes precedence over n°2. If sharing is enabled for a group but public sharing disabled for that same group, then the group will only be able to create internal shares.

lib/Model/Settings/AppSettings.php
dartcafe: This is what the frontend tells me. 

If `shareapi_exclude_groups === 'no'` the list should get ignored. Of course just if the mode is requested correctly. 

image

artonge:
> I mean: What is the difference between Exclude groups from creating link shares and Exclude some groups from sharing?

"`sharing`" means internal and external share.
"`external share`" and "`link share`" both means public link.

> If you want to say this block is all about public links, then the terms should get unified.

Yes, it is all about public link. Indeed, this could be harmonized.

lib/Model/Settings/AppSettings.php
```diff
@@ -132,6 +132,65 @@ public function getComboAllowed(): bool {
...
+
+		if ($groupExceptionMode === 'allowGroup') {
+			// exception mode is 'limit sharing to some groups'
+			// if user is in exception group, allow share creation
+			return $this->userSession->getUser()->getIsInGroupArray($excludedGroups);
+		}
+
+		// exception mode is 'Exclude some Groups from sharing'
+		// if user is in exception group, deny share creation
+		return $this->userSession->getUser()->getIsInGroupArray($excludedGroups);
```
hamza221, line 163:
```suggestion
		return !$this->userSession->getUser()->getIsInGroupArray($excludedGroups);
```

lib/Model/Settings/AppSettings.php
hamza221, line 186:
```suggestion
		$excludedMode = $this->config->getAppValue('core', 'shareapi_exclude_groups', '');
```

dartcafe: Hey @hamza221 Thanks for reviewing. But you may save time if you wait until I am finished with the next commit. If my discussion is solved, I will implement the final logic which will be ready for review and testing. 

Just want to avoid duplicated work.

dartcafe:
The we have to differ between the meanings. In Polls there are different types of links: 
* `public link` means a public accessible link to a poll, where (internal and external) users are able to register themself. 
  The result of the registration is an internal or external `personal link`
* `personal link` means a personalized share
  There are two different types:
  * `internal personal link` means a share exclusively for one registered Nextcloud user and only usable if the user is logged in
  * `external personal link` means a share exclusively created for an external user

And we have `open accessible polls`, which means all internal users are allowed to access the poll without any share.

Note: The creation of `public links` can be allowed or disallowed per group inside of Poll's settings.
image

As well as the permission to set polls as `openly accessible polls`.

In conclusion I understood that
* the setting `Limit sharing based on groups` should have effect on all of Poll's  shares:
  * `Allow sharing for everyone` allows all kind of shares, unless Poll's setting disallows `public shares`
  * `Exclude some groups from sharing` **allows** **all shares**, unless Poll's setting disallows `public shares`. 
  For the listed groups **all shares** including `public shares` are **disallowed**, regardless of Poll's setting for `public shares`.
  * `Limit sharing to some groups` **disallows** **all shares**, regardless what Poll's setting is set to. 
  For the listed groups all shares are **allowed**, except `public links` if Poll's setting disallows `public shares`
* The setting `Exclude groups from creating link shares` is ignored at the moment but should disallow `external personal links` and `public links`

Is my conclusion correct?

lib/Model/Settings/AppSettings.php
dartcafe: Different solution was implemented.

lib/Model/Settings/AppSettings.php
dartcafe: Already fixed and enhanced.

dartcafe:
Finished the backend regarding 'Limit sharing based on groups'
* Some janitor tasks
* remove sharing option in frontend, if sharing is disabled. 

To do: 
* Implement restriction of external shares

@hamza221 You may review meanwhile, if you are able to. My tests were positiv.

dartcafe:
I plan to release this change until this evening.

dartcafe:
Using this setting should now (hopefully) also avoid adding external shares and resolving circles to external shares:
grafik

hamza221 (approved):

icewind1991 (approved):

dartcafe:
Thanks guys!

PR #3915, opened 2025-03-15
fix typo in error message

Yannick-XY:


dartcafe:
Thanks, but the error is originated in transifex and has to be fixed there.

I will fix it there.

PR #3930, opened 2025-03-24
Votes rank feature enabled

dartcafe:


lib/Db/Poll.php
```diff
@@ -78,7 +80,8 @@
 class Poll extends EntityWithUser implements JsonSerializable {
 	public const TABLE = 'polls_polls';
 	public const TYPE_DATE = 'datePoll';
-	public const TYPE_TEXT = 'textPoll';
+	public const TYPE_IND_TEXT = 'textIndPoll';
+	public const TYPE_RANK_TEXT = 'textRankPoll';
```
dartcafe, line 84: I would prefer leaving the constants unchanged. Don't change the constants. Just add the new one (`TYPE_RANKED_TEXT`).

Or better `TYPE_TEXT_RANKED`

lib/Db/Poll.php
```diff
@@ -135,6 +138,7 @@ class Poll extends EntityWithUser implements JsonSerializable {
 	protected string $access = '';
 	protected int $anonymous = 0;
 	protected int $allowMaybe = 0;
+	protected $choosenRank = [];
```
dartcafe, line 141: Avoid array as database type. I am not sure, if it work with all db engines properly. 
Prefer a string and add getters which decode and encode the string into an array.
BTW: typo `choosen` => `chosen`

src/js/components/Configuration/ConfigRankOptions.vue
```diff
@@ -0,0 +1,171 @@
+<template>
+  <div class="option-container">
+    <!-- Menu déroulant pour sélectionner une option -->
```
dartcafe, line 3: Please. Comments always in English language!

dartcafe (commented):
First quick look.

lib/Db/Poll.php
vinimoz: if it doesnt mind u i've keep TYPE_RANKED_POOL like my field is textRankPool and as i'm lazy

lib/Db/Poll.php
vinimoz: ya right. I've change it to string i will test the modification.

vinimoz:
Sorry, but i'm a newbie, in github, vue, etc :smile: . I've proceed with modifications and fix.  I've try several time to put in the branch master-7, Rank votes features, without luck,  i always get access denied. I will put the update on my fork

vinimoz (approved):

vinimoz:
After several tries, i manage to do a commit with the changes. I can see my commit. On the commit. I was trying to change the mode draft but i think i cannot do that.

lib/Db/Poll.php
dartcafe: It is because of sorting and hierarchical context. At first it is a text poll and second there is a voting variation. 

That is not important, but changing the existing types is no good idea. Already existing polls are marked as textPoll inside the database. 

Leaving the existent constants name results in less code changes to review.

dartcafe:
> Sorry, but i'm a newbie, in github, vue, etc 😄 . I've proceed with modifications and fix. I've try several time to put in the branch master-7, Rank votes features, without luck, i always get access denied. I will put the update on my fork

Don't care. That's the way I went through, when I started here.

dartcafe:
> After several tries, i manage to do a commit with the changes. I can see my commit. On the commit. I was trying to change the mode draft but i think i cannot do that.

You can't change it, because you are not a member of the nextcloud organisation (organisation here on Github). All changes are done inside your branch and end here in this PR as soon as you commit them to your branch (in your repo)

dartcafe:
@vinimoz To make the changes better readable please run `composer run cs:fix`. 

You could also run eslint (`npm run lint:fix`)

vinimoz:
Yesterday, i commit and other on. eslint doenst give me any more error. Regarding phpcs and php-cs for lib directory
php-cs : Found 82 of 197 files that can be fixed in 1.444 seconds, 23.12 MB memory used
phpcs: PHPCBF CAN FIX THE 398 MARKED SNIFF VIOLATIONS AUTOMATICALLY
I don't put the fix options and commit it as i modify only 3 files in lib directory. If u want i do it an commit it.

vinimoz:
i think i commit in my fork https://github.com/vinimoz/polls/tree/pr3930_contrib

dartcafe:
something is not clean here. 

You commited reverted commits from the master-7 branch (affecting the translation files). How did you create your branch?

vinimoz:
Oups sorry, maybe it's my fault it coz someone else was working on my 
branch of the fork, Rank_votes_feature ?.

Yesterday, i try to commit on the pr3930, as i was seeing my commit 
there. After realize i haven't the right, after go back on the correct 
place, i do a mistake a create a branch 3930_commit on my main branch by 
error. So i after i merge it with Rank_vote s_feature and delete it the 
3930_commit. After i've tried to fight to fix all DCO.

To create my branch originally after the fork i do, after i remember, i 
clone it, fetch the upstream, i do gitcheckout -b Rank_votes_feature. 
Its what i remember.

dartcafe:
@vinimoz Are you still working on this? 

I would prefer to close this PR out of the following reasons: 
* This PR is based on v7 and v7 will slowly become legacy, since the v8 is nearly about to leave beta state
* v8's code base is not compatible to v7's, so any changes will have a short life
* I already prepared voting variants [here](https://github.com/nextcloud/polls/pull/3973)
  A ranked voting is rather a voting variant but a poll type. 
* We should think the implementation first from the UI. 
  I prefer another UI for ranked votes (something like moving options by drag and drop into an order instead of entering numbers)

vinimoz:
I’ve stopped working on it since, u told me something wrong happens,  
I had modified some files that ideally shouldn’t have been changed. That 
was my understanding up to that point.
That said, throughout the development, I made an effort to keep changes 
to a minimum and maintain the flexibility you had implemented for the 
other voting types. With the current setup, we can proceed with votes, 
smiley like, score voting, letter grades, majority judgment, Borda, 
Nauru, and Condorcet. I do agree that a drag-and-drop feature would 
improve usability, especially for Condorcet, and could definitely be 
refined.
Regarding the version, I was aware of the situation. That’s one of the 
reasons I aimed to deliver a usable version as quickly as possible to 
avoid blocking progress. Migrating to Vue 3 shouldn't be a major issue. 
Also, I needed those improvements fairly quickly for our political group.
I understand your point of view. What I don’t quite understand is why I 
was asked to fix or change things if, in the end, it wasn’t what you 
were looking for. It would’ve simply been more transparent and 
respectful to clarify that from the beginning.

Le 21/04/2025 à 06:01, René Gieling a écrit :

artlog:
@vinimoz i am pushing forward this with #4200 pull request to converge for this enhancement to see the light. PRocess still ongoing.

dartcafe:
@vinimoz  I think this is superseded by #4200. So I will close it here. 

And maybe you want to have a look at it, since @artlog adopted your work there.

PR #3938, opened 2025-03-26
fix: html and excel exports

### LINKED ISSUE #3933: Export wrong before page refresh for HTML and Excel spreadsheet

- [x] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [x] This issue is **not** already reported on Github _(I've searched it)_.

Create a poll, i.e. with for option, vote one time, ticking at least 1 box, export via top-right.

* CSV contains a record with 1 vote (1 checkmark, 3 Xs)
* HTML/Excel contain one record with no vote (4 Xs)

Hit refresh in the browser

* CSV, HTML and Excel export contain 1 vote (1 checkmark, 3 Xs)

Image

Export is correct for all formats without the need of a refresh

1. create a poll with 4 option
2. vote with ticking 1 box
3. export without page refresh for csv, html, excel

### Installation method

Installed/updated using occ

### Installation type

Updated from a minor version within same major version (i.e. 4.0.0 to 4.1.1)

### Affected polls version

7.2.5->7.4.1

- [x] Firefox
- [ ] Chrome
- [x] Chromium/Chromium based (i.e. Edge)
- [ ] Safari
- [ ] Other/Don't know

### Other browser

_No response_

### Add your browser log here

```json

```

_No response_

### NC version

Nextcloud 29

### Other Nextcloud version

_No response_

### PHP engine version

PHP 8.2

### Other PHP version

_No response_

### Database engine

MySQL

_No response_

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other/Don't know

```shell

```

_No response_

### Configuration report

```shell

```

### List of activated Apps

```shell

```

### Nextcloud Signing status

```shell

```

### Additional Information

_No response_

hamza221:
fix #3933

### DIFF
--- a/lib/Db/Vote.php
+++ b/lib/Db/Vote.php
@@ -58,6 +58,19 @@ class Vote extends EntityWithUser implements JsonSerializable {
 		$this->addType('deleted', 'integer');
 	}
 
+	public function getAnswerSymbol(): string {
+		switch ($this->getVoteAnswer()) {
+			case self::VOTE_YES:
+				return '✔';
+			case self::VOTE_NO:
+				return '❌';
+			case self::VOTE_EVENTUALLY:
+				return '❔';
+			default:
+				return '';
+		}
+	}
+
 	/**
 	 * @return array
 	 *
@@ -69,6 +82,7 @@ class Vote extends EntityWithUser implements JsonSerializable {
 			'pollId' => $this->getPollId(),
 			'optionText' => $this->getVoteOptionText(),
 			'answer' => $this->getVoteAnswer(),
+			'answerSymbol' => $this->getAnswerSymbol(),
 			'deleted' => $this->getDeleted(),
 			'optionId' => $this->getOptionId(),
 			'user' => $this->getUser(),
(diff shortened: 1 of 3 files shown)


dartcafe (commented):
I wonder, if it wouldn't be better to let the symbols get set by the backend.
I will push a commit as proposal.

Are there any caveats to expect, when sending symbols inside the REST response?

AndyScherzinger:
I'd expect that to be fine. Only thing I could think of is character-set issues but than we would experience problems with emojis through the apps / Talk as well. Hence I'd consider it save

dartcafe:
@hamza221 OK with you? Then you may merge, because it is your PR.

dartcafe (approved):

dartcafe:
Fun fact. I just wanted to do a manual backport and found that I already implemented this change [three months ago](https://github.com/nextcloud/polls/pull/3825/commits/c5d8ac7243ba2343cf2fac02d818d9fa8cdaf807#diff-549603f9983924f7dcc1eda99c170e0abaed533a81d6974e9153b500e95a47a2R83-R107). :rofl:

AndyScherzinger:
Which proves the approach if you were able to repeat it 🎉

PR #3958, opened 2025-04-04
Preparation for the update to @nextcloud-vue v9.0.0-alpha.8

dartcafe:
### :exclamation: This branch is still on alpha.7

* adopt changed props
* Datepickers are broken.


src/components/Actions/modules/ActionAddOption.vue
```diff
@@ -31,7 +31,7 @@ async function clickAction() {
 <template>
 	<div class="action option-date-add">
 		<NcButton
-			:type="ButtonType.Primary"
+			:variant="ButtonVariant.Primary"
```
ShGKme, line 34: Note: now you can use plain string like `<NcButton variant="primary" />`

Eusing exported object is not required anymore.

susnux: > Using exported object is not required anymore.

And deprecated as enums in general are a legacy concept in Typescript.
What will and did always work is to just use the string value so probably the exports will become type only exports of those strings.

dartcafe: Sad. I just like it, since I the have a dictionary of valid values inside the IDE. So you plan to remove them?

dartcafe: And for the prop migration it was a very helpful navigation to find all places, where type had to be replaced by variant. To be honest, it was a global search&replace

ShGKme: > Sad. I just like it, since I the have a dictionary of valid values inside the IDE. So you plan to remove them?

IDE still has the list of values (it comes from typing) and check for the correct value (from TS).

The difference is that you imported an object, added it to the component instance properties, and then used in the `<NcButton>`. Now all of it isn't needed and you can just use the value.

ShGKme: > And for the prop migration it was a very helpful navigation to find all places, where type had to be replaced by variant. To be honest, it was a global search&replace

It is still possible to find these places with `:?type="`.

You'd still have to check every result before replacing, but you needed it anyway.Using exported ENUMs didn't forbid using value from a variable.

Issue #3322, opened 2024-02-16
Three UI Proposals for polls with many participants: Sticky Option Headers, Sort/Group Participants by selected option, Show participants on hover


Bibo-Joshi:
### ⚠️ This issue respects the following points: ⚠️

- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Describe the goal you'd like to achieve

As a **poll participant _or_ owner**, I want to **easily get an overview about who has voted for what**, because **helps to select the accepted option (in case of a par, one participant might be more important) or to contact other people that voted for the same option**.

## Sticky Option Headers

In the table view, when scrolling through the list of participants, the column headers (the options) vanish and one easily looses track of which column is what. If the headers were sticky (like the poll name), that would be much easier

## Sort/Group Participants by selected option

Again in the table view, when I want to see all participants that voted for Option A, the only way to do that is to scroll through the complete list and check manually. Depending on the order in which the participants voted, only the first and last user could have voted for option A, all others for option B, making it necessary to scroll to the very bottom.
I propose that clicking on the Option text "A" sorts the participants like this:

1. current user
2. users that voted for option A, in the order of participation
3. users that voted "maybe" for option A, in the order of participation
4. users that did not vote for option A, in the order of participation

second click sorts "reversed", i.e. 2-4. are reversed. Third click resets to default sorting. A small triangle up/down icon can indicate the sorting.

##  Show participants on hover

In the list view, it's currently not possible at all to see who voted for what. I propose that hovering over the "voted" / "voted maybe" icons shows the profile pictures of the users that voted for that option. Possible solutions are a proper popup or displaying a scrollable list (including profile picture & name) or using the space between option text on the left and vote count on the right to display a (possibly trunacted) list of overlapping profile pictures.

I admit that showing the users in the list view is much harder and the two solutions I proposed both have different drawbacks. The two proposals for the table view are much more important to me.

### Additional context

For the sticky option headers, I found the discussion in #305, which is kind of similar, but proposed to make the current users votes sticky as well. I do not see a need for that, since (as a result of #305 ?!) it'l already at the very top.

---
Thanks for the work on this cool app!

Bibo-Joshi:
As suggested in https://github.com/nextcloud/polls/pull/3988#issuecomment-3011726162, I'll split this issue up. I'll link to the new ones in an edit of the comment in a minute.

Here we go: #4135, #4136, #4137

Issue #2993, opened 2023-07-13
Top-Scrollbar


Coernel82:
### ⚠️ This issue respects the following points: ⚠️

- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Describe the goal you'd like to achieve

# Add a top scrollbar
When viewing the results in grid view of larger polls horizontal scrolling is difficult. To make it easier add an additional scroll bar above the line with the dates.

dartcafe:
I don't understand your issue. What is difficult, why, under which circumstances. 

In the issue form, there is a sentence above the field which states: 

___
### Describe the goal you'd like to achieve
**Describe what (goal) you want to achieve and why (value).** Please explain your goal and why this is valuable and
**do not** describe the solution you suggest. Try to use the [user story](https://en.wikipedia.org/wiki/User_story) pattern"

>As a **[poll participant/poll owner/...]**, I want to **[be able to .../ don't want to ...]**, because **[it helps to ... /avoids ...]**.

___
Maybe, I could add a second scroll bar on the top of the table - **_which would look weird IMHO_** and would break any design pattern - but I am not aware, what problem will be soved then. 

A field for a solution description followes then.

Coernel82:
Sorry for the misunderstanding - I thought the issue would be so clear that no further explanation would be needed. And sorry for the late reply as I was on holidays.

As a poll owner I want to be able to scroll the result of larger polls horizontally without scrolling to the bottom to make the scrollbar visible. This sometimes makes sense if I want to get an overview over the poll results of a single user which is located at the top of the poll.

To give a practical example: I have the first user and want to see a date result which is at there very right end. To see this I need to scroll down and then scroll to the right end and after that scroll up again as the user is not visible any more.

So the solution could even be better - just getting the idea: make the scrollbars position fixed for polls which horizontally exceed the screen width. This would not look weird!

dartcafe:
Ah. ok. You say, tha scroll bar is out of sight, if there are more voters, than space to display them. OK. Got it. 

Scrolling is alway an ugly issue. I have to think about possible solutions.

PR #3988, opened 2025-04-26
Sort options by yes votes

dartcafe:
* migrate vote table mainly to grid view
* Add sorting of participants by yes-votes per option #3322
* Prepare table for sticky option headers (not functional yet) #3322, #2993

* moved removal of other user's votes to a menu connected to the avatars
* moved user options from the cog-wheel-menu to current user's menu connected to the avatar (only in table-view)


Bibo-Joshi:
Hi. Very cool to see the activity for 8.x! I just wanted to clarify:

> Add sorting of participants by yes-votes per option https://github.com/nextcloud/polls/issues/3322

This is not what I meant in that issue with "Sort/Group Participants by selected option" 🫠 
It seems that you can now sort the *options* such that options with a high number of yes-votes are shown first. This is ofc useful as well!
What I was aiming for however is sorting the users *within an option* such that the yes-voters are shown first before the maybe- & no-voters

dartcafe:
Did you see the three dots menu?  You will see a sort option:
image

Bibo-Joshi:
Aha, it's there on polls that I created (or got transferred into my ownership?) but the three dots are not available on polls that I'm just participating in. I tested on the latter yesterday, that's why I didn't see it. Awesome that the functionality is there now 🎊 thanks a lot! My use case would still require the sorting to be available to participants. Should I report that as a separate issue or should it just be tracked as part of #3322?

dartcafe:
Add it as another issue.
BTW: Never bundle multiple stories, they are not easy to track.

PR #3996, opened 2025-04-30
Refactor BasePublicController to reduce duplicated code

GeorgeApos:
While running a SonarQube scan on the project, I noticed that BasePublicController had 62.5% duplicated lines, primarily due to repeated try/catch logic in the response(), responseCreate(), and responseLong() methods.

This PR refactors the class by introducing a single private method, handleResponse(), which centralizes the core response logic while remaining flexible for different status codes and exception types.

The result is cleaner, more maintainable code that remains functionally identical.


dartcafe:
Nice. One item of the I-have-to-look-after-that-later-list. Do you think, you could the other base controllers as well? 

And you should sign off the commits.

lib/Controller/BasePublicController.php
```diff
@@ -47,22 +43,38 @@ protected function response(Closure $callback): JSONResponse {
...
+    		} catch (\Throwable $e) {
+    			if (is_a($e, $exceptionClass, true)) {
+    				if ($fallbackStatus !== null && $e instanceof NoUpdatesException) {
+    					return new JSONResponse([], $fallbackStatus);
+    				}
+    				if ($e instanceof Exception) {
+    					return new JSONResponse(['message' => $e->getMessage()], $e->getStatus());
+    				}
+    			}
+    			return new JSONResponse(['message' => 'Unexpected error'], Http::STATUS_INTERNAL_SERVER_ERROR);
```
dartcafe, line 74: I am unsure, but I think 5xx is handled by the server itself and this way we would lose the trace log.

GeorgeApos: You're right that catching \Throwable and returning a custom 5xx response might suppress the server's own error handling and logging. To avoid losing the trace log, I could either rethrow the exception after logging it, or limit the catch block to Exception (not \Throwable) and let fatal errors bubble up as usual.

dartcafe: Better rethrow it IMHO, so this code is fine as it is.

GeorgeApos:
> Nice. One item of the I-have-to-look-after-that-later-list. Do you think, you could the other base controllers as well?
> 
> And you should sign off the commits.

I’ve refactored the other base controllers in the same way and signed off the commits.

lib/Controller/BaseApiV1Controller.php
```diff
@@ -63,10 +56,27 @@ protected function responseLong(Closure $callback): JSONResponse {
...
+		int $successStatus = Http::STATUS_OK,
+		bool $checkNoUpdates = false
+	): JSONResponse {
 		try {
-			return new JSONResponse($callback(), Http::STATUS_CREATED);
-		} catch (Exception $e) {
-			return new JSONResponse(['message' => $e->getMessage()], $e->getStatus());
+			return new JSONResponse($callback(), $successStatus);
+		} catch (\Throwable $e) {
+			if ($checkNoUpdates && $e instanceof NoUpdatesException) {
```
dartcafe, line 70: $checkNoUpdates seems to be not necessary to me, since NoUpdatesException is only thrown for long polls

lib/Controller/BaseApiV2Controller.php
```diff
@@ -70,10 +58,27 @@ protected function responseLong(Closure $callback): DataResponse {
...
+		Closure $callback,
+		int $successStatus = Http::STATUS_OK,
+		bool $checkNoUpdates = false
+	): DataResponse {
 		try {
-			return new DataResponse($callback(), Http::STATUS_CREATED);
-		} catch (Exception $e) {
-			throw new OCSBadRequestException($e->getMessage());
+			return new DataResponse($callback(), $successStatus);
+		} catch (\Throwable $e) {
```
dartcafe, line 71: No \Throwable

lib/Controller/BaseApiV2Controller.php
```diff
@@ -70,10 +58,27 @@ protected function responseLong(Closure $callback): DataResponse {
...
+			if ($checkNoUpdates && $e instanceof NoUpdatesException) {
```
dartcafe, line 72: $checkNoUpdates not necessary... ?!

lib/Controller/BaseApiV1Controller.php
```diff
@@ -63,10 +56,27 @@ protected function responseLong(Closure $callback): JSONResponse {
...
+				return new JSONResponse([], Http::STATUS_NOT_MODIFIED);
+			}
+			if ($e instanceof DoesNotExistException) {
+				return new JSONResponse(['message' => $e->getMessage()], Http::STATUS_NOT_FOUND);
+			}
+			if ($e instanceof Exception) {
+				return new JSONResponse(['message' => $e->getMessage()], Http::STATUS_INTERNAL_SERVER_ERROR);
+			}
+			return new JSONResponse(['message' => $e->getMessage()], Http::STATUS_INTERNAL_SERVER_ERROR);
```
dartcafe, line 79: re-throw $e here.

lib/Controller/BaseController.php
```diff
@@ -74,12 +62,42 @@ protected function responseCreate(Closure $callback): JSONResponse {
...
+		string $secondaryException = null
+	): JSONResponse {
 		try {
-			return new JSONResponse($callback());
-		} catch (DoesNotExistException $e) {
-			return new JSONResponse(['message' => 'Not found, assume already deleted']);
-		} catch (Exception $e) {
-			return new JSONResponse(['message' => $e->getMessage()], $e->getStatus());
+			return new JSONResponse($callback(), $successStatus);
+		} catch (\Throwable $e) {
```
dartcafe, line 85: No \Throwable

lib/Controller/BasePublicController.php
```diff
@@ -47,22 +43,38 @@ protected function response(Closure $callback): JSONResponse {
...
+
+	private function handleResponse(
+    		Closure $callback,
+    		int $successStatus,
+    		string $exceptionClass,
+    		int $fallbackStatus = null
+    	): JSONResponse {
+    		try {
+    			return new JSONResponse($callback(), $successStatus);
+    		} catch (\Throwable $e) {
```
dartcafe, line 65: No \Throwable

GeorgeApos:
I made the changes you suggested. Apologies in advance if anything isnt aligned with your expectations, this is my first open-source contribution

dartcafe:
@GeorgeApos 

> Apologies in advance if anything isnt aligned with your expectations, this is my first open-source contribution

:laughing: Don't care. There is no need for excuse. It is the way it is done; Everything is perfect.

But sorry, I hijacked your PR, but while reviewing it, my ambitions raised to optimize it further and it was more straight forward to implement rather than explain.

What I did: 
* Reduce all response calls to a single method per controller
  * Detect a long poll by the exception status (304)
  * Eliminate `BaseController::responseDeleteTolerant()` by defining the new Exception `OCA\Polls\Exceptions\AlreadyDeletedException` using 304 as Status as well
  * use your attempt to define the `successCode` but moved it to the Controllers as a parameter and delete `responseCreate`; Set 200 as default

As result:
* eliminate the `BasePublicController` and use the `BaseController` as parent class for the public controllers since there was no more difference

Further: 
* Catch `OCA\Polls\Exceptions\Exception` instead of `\Exception`
* remove `BaseApiV1Controller`, as it is no more used
* removed leftover catching of `\Throwable`

As the result the controller should behave as before, but with much more leaner code. You had a nice attempt :+1: 

Tell me what you think of the changes, I made.

dartcafe:
Hey @GeorgeApos 

Sorry, I don't want to rush you, you may have things to do for sure. But I am going to merge, because I want to avoid further merge conflicts.

Any occurring issues may be solved in a different fix. I reviewed the code several time, so I do not expect any problems.

GeorgeApos:
> Hey @GeorgeApos 
> 
> 
> 
> Sorry, I don't want to rush you, you may have things to do for sure. But I am going to merge, because I want to avoid further merge conflicts.
> 
> 
> 
> Any occurring issues may be solved in a different fix. I reviewed the code several time, so I do not expect any problems. 
> 
> 
> 
> And thank you for your support. That helped a lot. Any participation is welcome.

Hello, yeah of course, I saw the code, it turns out when you know the code base, you know how to make it perfect! Thank you for the merge and your help, if I find more problems I will point them out!

PR #4139, opened 2025-06-30
fix(l10n): apply translation guidelines

hamza221:
This fix is intended to bring the sources to respect the translation guidelines [Dos and Don'ts](https://docs.nextcloud.com/server/latest/developer_manual/basics/translations.html#dos-and-don-ts)


hamza221:
@dartcafe  I want to backport down to Nextcloud 30, that would be main and Master-7 right ?

dartcafe (approved):

dartcafe:
A backport to master-7 will break existing translations, because only main is handled by Transifex. 

Anyways 8.x will be released for Nextcloud 30 soon.

PR #4149, opened 2025-07-06
variant of fixed vote table header

dartcafe:
* make fixed vote header default
* add a variant with flex items (activate by user settings)
  In the result only the table is scrollable.
  Side effect is a narrow vote table on small screens and when a lot of info cards are used

with info cards | without info cards
-|-
grafik | grafik
grafik | grafik


dartcafe:
Remove from milestone, because it got an overwork and the variant got removed

PR #4352, opened 2025-10-25
Update API documentation for option addition

KnudH:
Fixing examples for adding options with and without sequences.


dartcafe:
Sorry for being late but I am currently rebuilding a flat and moving.

I did not test it, but I guess it works. Thanks for taking care here.

PR #4437, opened 2026-01-10
Change option's time handling and add timezones

### LINKED ISSUE #2368: Allow ignoring timezone for date only polls
## Describe the goal you'd like to achieve

As a poll participant I want to be able to view date only polls as such even if my timezone differs from the timezone of the creator, because it makes reading and filling out such a poll more easy. This is a poll where such feature could be useful:
image

## Describe possible solutions
Add a setting, which will allow ignoring the timezone of a poll if such poll only consists of full days. If such setting is enabled, the example poll from above should look like this (ignoring the change of language):
image

Another solution could be that the poll app allows configuring the timezone I want to use, which then will further ignore the timezone of the browser.

## Additional context
My browser sends the timezone UTC to all websites for privacy reasons, including Nextcloud. This is most of the time usable as good websites will list the timezone they are using to display the time so I know that. In the Nextcloud calendar, I configured my correct timezone so that the calendar will use my home timezone. However the creator of the poll has this setting not enabled and lives currently in a UTC+2 zone. If he/she/it creates the poll without inserting any times. I see the poll converted to my timezone which makes it really hard to read, because then a poll answer which reflects "Tuesday" will be displayed as a "Monday to Tuesday" answer, which I often interpret as "Monday".

### LINKED ISSUE #2851: Guests/Users can select timezone upon opening date+time polls

- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Describe the goal you'd like to achieve

Allow guest users and users to view times for time-based polls in their local timezone.

### Describe possible solutions

_No response_

### Additional context

_No response_

### LINKED ISSUE #3699: Support date-only poll options

- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Describe the goal you'd like to achieve

As a poll owner, I want to be able to suggest date-only options (without timing) because it allows for flexible scheduling, e.g. agreeing to meet in the evening on a certain date without agreeing to the exact time.

This was requested earlier here: https://github.com/nextcloud/polls/issues/2290

Many thanks for a great product!! 

### Describe possible solutions

_No response_

### Additional context

The issue today is that an all day event is displayed as bridging two different dates, which is confusing, because it looks like a two-day event is proposed. I hope the attached screenshots of poll visitor (first screenshot) and poll creator (second screenshot) display how unintuitive this is currently.

Screenshot_20240908-150832_Firefox Focus_1
Screenshot_20240908-150857_Mull_1

### LINKED ISSUE #3700: Allow user selection of poll time zone

- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Describe the goal you'd like to achieve

As a poll owner, I want to be able to manually set the time zone of a poll, because then I am certain which time zone the poll really uses.

### Describe possible solutions

Currently, I have to guess which timezone is used. If I use a privacy browser, then it sets to UTC. If I use my work laptop, it sets it to UTC+1 or UTC+2. If I use someone else's laptop, it uses some third time. And I cannot see this when making the poll. I have to login with a laptop/browser that I know has the intended time zone setting to verify what was set.

### Additional context

Please as this is implemented, do not automatically change the scheduled times if the time zone is changed by the poll owner.

Thank you for a great product, I use polls all the time ❤️

### LINKED ISSUE #4045: Group: Time zone issues

### Describe the goal you'd like to achieve

Grouped issue for time zone related issues

### Describe possible solutions

_No response_

### Additional context

_No response_

dartcafe:
* switch timestamp to ISO date
  may be reverted
* switch duraton in seconds to ISO Interval
* Allow to change default timezone per poll
* allow to select the preferred timezone for participants

closes #4045
closes #3700
closes #2368
closes #2851
closes #3699

Timezone selection is only available in date polls and if the polls's original timezone does NOT match the local timezone.

set timezone (poll configuration) | select original timezone set by owner | select visitor's local timezone | add option
-|-|-|-
grafik| grafik  | grafik | grafik

To dos:

- [x] Fix design issues
- [x] Fix layout issues with changing view ports
- [x] Code tidy
- [x] Backend tests
- [x] migration of options and votes

Affects options and votes:
Usage of timestamps have been replaced by DateTimeImmutable.
Duration in seconds has been replaced by DateInterval.

To avoid migrations now, the db will still receive timestamps and duration values in seconds, but also ISO formatted strings representing the option's timestamp and duration.  API got extended by the ISO values.

ISO always values get precedence over their numeric sibling, either from the values read from the database or when set using the API.

### DIFF
--- a/lib/Db/Option.php
+++ b/lib/Db/Option.php
@@ -189,7 +189,6 @@ class Option extends EntityWithUser implements JsonSerializable {
 	 */
 	public function getOwnerUser(): ?UserBase {
 		if ($this->getUserId() === '') {
-			$this->getUser();
 			return null;
 		}
 		return parent::getUser();
@@ -308,7 +307,6 @@ class Option extends EntityWithUser implements JsonSerializable {
 		$this->setOrder($dateTime->getTimestamp());
 		/** @psalm-suppress UndefinedMagicMethod */
 		$this->setIsoTimestamp($dateTime->getISO());
-		$this->setOrder($dateTime->getTimestamp());
 		$this->setText($this->getDateTimePollOptionText());
 	}
 
--- a/lib/Model/SimpleOption.php
+++ b/lib/Model/SimpleOption.php
@@ -24,6 +24,11 @@ class SimpleOption implements JsonSerializable {
 	protected DateTimeImmutable $dateTime;
 	protected DateInterval $interval;
 
+	public function __construct() {
+		$this->dateTime = new DateTimeImmutable(null);
+		$this->interval = new DateInterval(null);
+	}
+
 	public function jsonSerialize(): array {
 		return [
 			'text' => $this->getText(),
@@ -65,7 +70,7 @@ class SimpleOption implements JsonSerializable {
 	}
 
 	/**
-	 * Get the timestamp of the option as DateTimeImmutable or null if not set
+	 * Get the timestamp of the option as DateTimeImmutable
 	 * @return DateTimeImmutable
 	 */
 	public function getDateTime(): DateTimeImmutable {
@@ -73,7 +78,7 @@ class SimpleOption implements JsonSerializable {
 	}
 
 	/**
-	 * Get the duration of the option as DateInterval or null if not set
+	 * Get the duration of the option as DateInterval
 	 * @return DateInterval
 	 */
 	public function getInterval(): DateInterval {
--- a/lib/Service/OptionService.php
+++ b/lib/Service/OptionService.php
@@ -124,10 +124,6 @@ class OptionService {
 		if (!$this->poll->getIsPollOwner()) {
 			$newOption->setOwner($this->userSession->getCurrentUserId());
 		}
-		$this->logger->error('Trying to add option', [
-			'optionText' => json_encode($newOption),
-			'pollId' => $pollId,
-		]);
 		try {
 			// Insert the new option
 			$newOption = $this->optionMapper->insert($newOption);
@@ -136,13 +132,12 @@ class OptionService {
 				'optionText' => json_encode($newOption),
 				'pollId' => $pollId,
 				'exceptionMessage' => $e->getMessage(),
-				'exceptionReason' => $e->getReason(),
 			]);
 			// TODO: Change exception catch to actual exception
 			// Currently OC\DB\Exceptions\DbalException is thrown instead of
 			// UniqueConstraintViolationException
 			// since the exception is from private namespace, we check the type string
-			if (get_class($e) === 'OC\DB\Exceptions\DbalException' || $e->getReason() === Exception::REASON_UNIQUE_CONSTRAINT_VIOLATION) {
+			if (get_class($e) === 'OC\DB\Exceptions\DbalException' && $e->getReason() === Exception::REASON_UNIQUE_CONSTRAINT_VIOLATION) {
 
 				// Option already exists, so we need to update the existing one
 				// and remove deleted setting
--- a/src/Api/modules/api.types.ts
+++ b/src/Api/modules/api.types.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import type { Option } from '../../stores/options.types'
+import type { OptionDto } from '../../stores/options.types'
 import type { Vote } from '../../stores/votes.types'
 import type { Poll } from '../../stores/poll.types'
 import type { Comment } from '../../stores/comments.types'
@@ -28,7 +28,7 @@ export type ApiEmailAdressList = {
 
 export type FullPollResponse = {
 	poll: Poll
-	options: Option[]
+	options: OptionDto[]
 	votes: Vote[]
 	comments: Comment[]
 	shares: Share[]
@@ -36,21 +36,21 @@ export type FullPollResponse = {
 }
 
 export type AddOptionResponse = {
-	option: Option
-	repetitions: Option[]
-	options: Option[]
+	option: OptionDto
+	repetitions: OptionDto[]
+	options: OptionDto[]
 	votes: Vote[]
 }
 
 export type setVoteResponse = {
 	vote: Vote
 	poll: Poll
-	options: Option[]
+	options: OptionDto[]
 	votes: Vote[]
 }
 
 export type RemoveVotesResponse = {
 	poll: Poll
-	options: Option[]
+	options: OptionDto[]
 	votes: Vote[]
 }
--- a/src/components/Actions/modules/ActionAddPoll.vue
+++ b/src/components/Actions/modules/ActionAddPoll.vue
@@ -22,7 +22,7 @@ import type { ButtonVariant } from '@nextcloud/vue/components/NcButton'
 
 interface Props {
 	caption?: string
-	modalSize?: string
+	modalSize?: 'small' | 'normal' | 'full' | 'large'
 	buttonMode?: ButtonMode
 }
 
--- a/src/components/Base/modules/ButtonModal.vue
+++ b/src/components/Base/modules/ButtonModal.vue
@@ -18,7 +18,7 @@ interface Props {
 	buttonVariant?: ButtonVariant
 	buttonMode?: ButtonMode
 	buttonCaption?: string
-	modalSize?: string
+	modalSize?: 'small' | 'normal' | 'full' | 'large'
 	noClose?: boolean
 }
 
--- a/src/components/Base/modules/DateBox.vue
+++ b/src/components/Base/modules/DateBox.vue
@@ -11,7 +11,7 @@ import { useSessionStore } from '@/stores/session'
 
 interface Props {
 	startDate: DateTime
-	duration?: Duration
+	duration: Duration
 	timezone?: string | undefined
 }
 
--- a/src/components/Options/OptionCloneDate.vue
+++ b/src/components/Options/OptionCloneDate.vue
@@ -36,7 +36,7 @@ const sequence = ref<Sequence>({
 })
 
 const dateBaseOptionString = computed(() =>
-	DateTime.fromISO(option.isoTimestamp).toLocaleString(DateTime.DATETIME_FULL),
+	option.getDateTime().toLocaleString(DateTime.DATETIME_FULL),
 )
 
 /**
--- a/src/components/Options/OptionItem.vue
+++ b/src/components/Options/OptionItem.vue
@@ -9,7 +9,6 @@ import DragIcon from 'vue-material-design-icons/DotsVertical.vue'
 import DateBox from '../Base/modules/DateBox.vue'
 import { usePollStore } from '../../stores/poll'
 import OptionItemOwner from './OptionItemOwner.vue'
-import { DateTime, Duration } from 'luxon'
 
 import type { Option } from '../../stores/options.types'
 
@@ -57,8 +56,8 @@ const pollStore = usePollStore()
 		<DateBox
 			v-else
 			class="option-item__option--date"
-			:start-date="DateTime.fromISO(option.isoTimestamp)"
-			:duration="Duration.fromISO(option.isoDuration)" />
+			:start-date="option.getDateTime()"
+			:duration="option.getDuration()" />
 
 		<slot name="actions" />
 	</component>
--- a/src/components/Options/OptionsDateAddDialog.vue
+++ b/src/components/Options/OptionsDateAddDialog.vue
@@ -88,20 +88,23 @@ const startDateTime = ref(
 )
 
 /**
- * Resultin option based on the inputs
+ * Resulting option based on the inputs
  */
+
 const newOption = computed<SimpleOption>(() => {
 	const dateTime = startDateTime.value
 		.startOf(allDay.value ? 'day' : 'minute')
 		.setZone(sessionStore.currentTimezoneName, { keepLocalTime: true })
 
-	return {
+	const dur = duration.value
+	const option: SimpleOption = {
 		text: '',
-		duration: duration.value.as('seconds') || 0,
-		timestamp: dateTime.toUnixInteger(),
 		isoTimestamp: dateTime.toISO() || '',
-		isoDuration: duration.value.toISO() || 'PT0S',
+		isoDuration: dur.toISO() || 'PT0S',
+		getDateTime: () => dateTime,
+		getDuration: () => dur,
 	}
+	return option
 })
 
 /**
--- a/src/components/Options/OptionsPreviewBox.vue
+++ b/src/components/Options/OptionsPreviewBox.vue
@@ -30,15 +30,9 @@ const {
 	title,
 } = defineProps<Props>()
 
-const dateTime = computed(() =>
-	DateTime.fromISO(option.isoTimestamp).setZone(timezone),
-)
-
-const duration = computed(() => Duration.fromISO(option.isoDuration || 'PT0S'))
-
 // computed last from dateTime repetition
 const lastRepetitionDateTime = computed(() =>
-	dateTime.value.plus(sequenceDuration.value),
+	option.getDateTime().plus(sequenceDuration.value),
 )
 
 const sequenceDuration = computed(() =>
@@ -56,7 +50,7 @@ const repetitionCaption = computed(() =>
 )
 
 const simpleDate = computed(() => {
-	const dates = getDates(dateTime.value, duration.value, otherTimeZone)
+	const dates = getDates(option.getDateTime(), option.getDuration(), otherTimeZone)
 	return dates.interval.toLocaleString(DateTime.DATETIME_MED)
 })
 </script>
@@ -70,8 +64,8 @@ const simpleDate = computed(() => {
 		</div>
 		<div class="preview-date">
 			<DateBox
-				:start-date="dateTime"
-				:duration="duration"
+				:start-date="option.getDateTime()"
+				:duration="option.getDuration()"
 				:timezone="timezone" />
 		</div>
 
@@ -79,7 +73,7 @@ const simpleDate = computed(() => {
 			<span>{{ repetitionCaption }}</span>
 			<DateBox
 				:start-date="lastRepetitionDateTime"
-				:duration="duration"
+				:duration="option.getDuration()"
 				:timezone="timezone" />
 		</div>
 		<div
--- a/src/components/Options/OptionsText.vue
+++ b/src/components/Options/OptionsText.vue
@@ -18,7 +18,8 @@ const pollStore = usePollStore()
 const optionsStore = useOptionsStore()
 // const element = useTemplateRef<HTMLElement>('list')
 
-useSortable(useTemplateRef<HTMLElement>('list'), optionsStore.options, {
+const listRef = useTemplateRef<HTMLElement>('list')
+useSortable(listRef, optionsStore.options, {
 	animation: 200,
 	group: 'options',
 	disabled: false,
@@ -30,7 +31,7 @@ useSortable(useTemplateRef<HTMLElement>('list'), optionsStore.options, {
 			/* do nothing, wait for nextTick() */
 		})
 	},
-})
+} as Parameters<typeof useSortable>[2])
 
 const componentStyle = {
 	'--content-deleted': `" (${t('polls', 'deleted')})"`,
--- a/src/composables/optionDateTime.ts
+++ b/src/composables/optionDateTime.ts
@@ -5,11 +5,10 @@
 
 import { Option } from '@/stores/options.types'
 import { DateTime, Duration, Interval } from 'luxon'
-import { computed } from 'vue'
 
 type OptionDateTime = {
 	optionStart: DateTime
-	duration: Duration | null
+	duration: Duration
 	optionEnd: DateTime
 	isFullDays: boolean
 	isSameMonth: boolean
@@ -22,7 +21,7 @@ export function getDatesFromOption(
 	option: Option,
 	timezone: string = Intl.DateTimeFormat().resolvedOptions().timeZone,
 ): OptionDateTime {
-	return getDates(option.isoTimestamp, option.isoDuration || null, timezone)
+	return getDates(option.getDateTime(), option.getDuration(), timezone)
 }
 /**
  *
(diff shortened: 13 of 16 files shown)


andreas-1107:
This is definitely a big improvement, thank you! One comment: 

I would recommend that the guest/visitor could choose to view it in the poll creator's / original timezone. This will remove confusion as to what was the intent of the poll creator, which I think is the main issue today. 

The behaviour seems to be the opposite in the screenshots. I understand New York is the poll creator's time zone, Berlin is the guest's time zone. I would suggest to switch the behaviour so that if I, as guest, click "original time zone" I see the poll as the creator made it (e.g. as if I am in New York). So I would see "23 January" only when selecting "original time zone". And if I switch to Berlin, I would see the Berlin times. 



A different note on UX, perhaps for next change: 

Other tools, for example Doodle, just ignores the time zone for full day events. If I create an all-day event in UTC 0 (Iceland), then it does not matter what time zone I select as a guest, it always displays the full day event as if I was in Iceland. Check here: https://doodle.com/group-poll/participate/e50qBgXd/vote

But time specific events are moved to the corresponding date. Screenshots below.

Thanks again for the good work 👍

Screenshot_20260111-104700_DuckDuckGo
Screenshot_20260111-104707_DuckDuckGo

PhrozenByte:
Great work, thanks for implementing this :heart:

I haven't really used polls for quite some time, but since I participated in the discussions intensively back then I felt the obligation to at least give it a try. The PR code runs without any issue (but I didn't do a code review), here are a few more suggestions (which can also be ignored since I'm basically a non-user):

* For all day events `polls` IMHO shouldn't even store time and timezone, but just the date (which should be possible with datetime strings now, e.g., `iso_timestamp = "2026-01-13"` and `iso_duration = ""`, or something like that, but I didn't check the code, just the database contents). For all-day events `polls` can't reasonably assume that the user truly wanted to create a 00:00-24:00 event, see https://github.com/nextcloud/polls/issues/2368#issuecomment-1103981919. `polls` should then just show the event with a date, but without time, which is also unaffected by whatever timezone the user selects
* Nevertheless it should still be possible to explicitly create 00:00-24:00 events (i.e., by creating an event that starts at 00:00 and takes 24 hours or 1 day, e.g., `iso_timestamp = "2026-01-13T00:00:00.000+01:00"` and `iso_duration = "P1D"`)
* I'm not sure whether `poll_option_text` is used for anything (seems like it's used when exporting a poll, but I'm not sure), but its datetime strings seem to always use UTC (or the server's timezone?), not the poll's timezone
* In general, `polls` IMHO should use the poll's timezone by default (basically just pre-select that in the timezone notice; the notice is hard to miss IMO)
* When creating new events as a poll owner, `polls` doesn't use the timezone selected in the timezone notice, but always the user's timezone (i.e., if I select the poll's timezone in the timezone notice and create new events, it should use the poll's timezone; if I select my local timezone instead, it should use my local timezone for new events)

dartcafe:
@andreas-1107 
> So I would see "23 January" only when selecting "original time zone". And if I switch to Berlin, I would see the Berlin times.

That is, how it works. In the example, New York ist the poll's timezone, defined by the owner. Berlin is the visitors (local) timezone. The pictures may be confusing due to the shifted dates of the "original" timezone.

I simply did not adjust the dates so that they match in a plausible way.

> Other tools, for example Doodle, just ignores the time zone for full day events. If I create an all-day event in UTC 0 (Iceland), then it does not matter what time zone I select as a guest, it always displays the full day event as if I was in Iceland.

I was thinking about it but it may have some issues, if the visitor sits far away (i.e. 12 timezones away). If the date is i.e. 2026-01-18, what does that mean for the visitor? Which day may be meant from his perspective. 2026-01-17, 2026-01-18 or 2026-01-19, depending of the direction (+12 or -12 hours)? This may be more confusing.

Most of the times the polls will show no timezone information, because owner and participants will reside within the same timezone.

I believe, that selecting the timezone in case of different timezones between visitor and owner is the most transparent information.

And to be honest, adding options with a particular timezone is still not done yet.

dartcafe:
@PhrozenByte 

> I haven't really used polls for quite some time, but since I participated in the discussions intensively back then I felt the obligation to at least give it a try. The PR code runs without any issue

:pray: Thanks for taking the time for testing. Be aware, that adding options inside a particulair timezone is still WIP.

> For all day events polls IMHO shouldn't even store time and timezone, but just the date (which should be possible with datetime strings now, e.g., iso_timestamp = "2026-01-13" and iso_duration = ""

Yes I am thinking about something like this, and try to fiddle out a good and working method. What keeps me from this for the moment is the migration. We have to be very very careful about it to avoid breaking existing polls. Currently the backend is not fully clean and needs a lot of polish. Fallbacks are needed as removing obsolete code. 

> Nevertheless it should still be possible to explicitly create 00:00-24:00 events

I guess this will be possible out of the box after the PR is finished.

> I'm not sure whether poll_option_text

I stores the options for a textpoll and a readable ISO like interval. Additionally it is a key for indices and for matching votes to an option. And yes, it is UTC here.

> In general, polls IMHO should use the poll's timezone by default (basically just pre-select that in the timezone notice; the notice is hard to miss IMO)

I plan to make this optional, so a poll owner can decide, which timezone should be used as default for a visitor.

> When creating new events as a poll owner, polls doesn't use the timezone selected in the timezone notice, but always the user's timezone (i.e., if I select the poll's timezone in the timezone notice and create new events, it should use the poll's timezone; if I select my local timezone instead, it should use my local timezone for new events)

Yes. Currently WIP, see comments before. The UI needs some tries to make it usable. In the end, both timezones will be displayed and also the timezone for the date-time selector should be selectable (Proposals of visitors in mind).

dartcafe:
@PhrozenByte @andreas-1107 

Thank you guys for this immediate responses. Helps a lot.

andreas-1107:
> > Other tools, for example Doodle, just ignores the time zone for full day events. If I create an all-day event in UTC 0 (Iceland), then it does not matter what time zone I select as a guest, it always displays the full day event as if I was in Iceland.
> 
> I was thinking about it but it may have some issues, if the visitor sits far away (i.e. 12 timezones away). If the date is i.e. 2026-01-18, what does that mean for the visitor? Which day may be meant from his perspective. 2026-01-17, 2026-01-18 or 2026-01-19, depending of the direction (+12 or -12 hours)? This may be more confusing.
> 
> Most of the times the polls will show no timezone information, because owner and participants will reside within the same timezone.
> 
> I believe, that selecting the timezone in case of different timezones between visitor and owner is the most transparent information.

I understand what you mean, and I think the case of +12 / -12 hours is sort of an edge case, as we tend to have meetings more rarely with people who are further away - and +/- 12h is always the furthest away it's possible to get :) 

I believe the full-day event is used most often when time zone differences are smaller (e.g. 1h or 2h), but even 1h difference in time zone causes confusion with the current UX as the event then appears to be spanning multiple dates.

So I think this is why Doodle have opted for keeping full-day events without time, and letting the user figure out which date is referred to using the context of the poll if there is any doubt (who created it, for example). As in the absolute majority of the cases, it will be obvious to the user what date is meant.

My 2 cents on this point is to allow the potential confusion of the +/- 12h time difference and prioritise the more common user experience of scheduling full-day events.

dartcafe:
The timezone difference between Moscow and New York counts 8 hours. Now imagine Trump wants to meet Putin and they miss each other because of misinterpreting the day. :wink: 

I think I understand the issue completely. I am absolutely sure, the majority of meeting organization happens within the same timezone. But I want to make sure, this works under any circumstances. 

This means for a whole day event: 
* either we mark ist as a 24h span and accept the shifts and allow to switch the timezone 
* or we just point to the date without any time information but have to add the timezone, where day fits. And of course we allow changing timezones also. 

The first solution will fit into the current scheme of Polls, how it handles dates and times. And it fits to exiting polls and just extends but does not change them. The second one can be a further optimization, but needs more time and coding and testing and migration.

BTW: With the planned option of setting the poll owner's timezone as the default should avoid the confusion too at first hand. 

And also thanks for participation in the discussion and adding your feedback. Valuable!!!

andreas-1107:
Ah, thank you for pointing out that the poll creator's time zone will be the default one (and not the visitor's browser for example). That will resolve the current UX confusion I think, as (I believe) most people won't change the time zone for polls with only full day events. 👍👍

dartcafe:
Adding options to a poll with flexible timezones was pushed. 
To dos are in the first comment.

dartcafe:
This took more time than expected, but I hope finally it is done. 
* Timezones are applied as I expected (in my tests).
* Internally date and time information are now handled by DateTimeImmutable class instead of a timestamp and
* durations are no more defined by seconds but by a time interval using ISO strings by the back-end.
* Timestamps and duration in seconds are still supported to avoid expensive migrations for the database and for the API.
* The API (option endpoints) got an extension, so that ISO strings are also supported.

This had more side effects and caveats than I was ready for. :grimacing: 

If anyone will give the beta a try and wants to roll back afterwards, just install the last release and 
* delete the columns `iso_duration` and 
* `iso_timestamp` from `oc_polls_options` and
* delete the migration entry from `oc_migrations`: app: `polls`, version: `090000Date20260302211900`

So @andreas-1107 and @PhrozenByte or anyone else: If you could give it a try and smoke test the beta (from the appstore or from the releases page) ...

v9.0.0-beta.2 will be out today. 

Since Nextcloud 31 is out of maintenance, this feature will mark the 9.0.0 milestone and Support of PHP 8.1 is also dropped. Although not explicitly tested, PHP 8.5 should work, too.

This PR will get merged soon, so please report any issues in a new issue.

PhrozenByte:
Great work, thanks @dartcafe! :heart: 

I tested with the current PR (fresh install from sources) and noticed the following 8 things (some suggestions, some bugs, and one note). I wasn't sure how to report this, so I decided to tell it like a step-by-step reproducer, I hope you can follow what I was doing:

---

**Step 1**

Let's assume I'm in "Europe/Berlin" and want to create a poll with "America/New York". I thus select "America/New York" as the poll's default timezone first. I then open the options modal. The defaults are: "Whole day" enabled, "2026-03-03" (i.e., today), timezone "Europe/Berlin" and a duration of "0 days". `polls` previews the following:

Screenshot 2026-03-03 at 18-04-14 Umfragen – Test

*1. Suggestion:* I'd recommend selecting the poll's default timezone by default here instead of the user's timezone.

*2. Suggestion:* I'd recommend using "1 days" as default duration.

Adding that option creates a `2026-03-02T23:00:00+00:00` option with `1D` duration.

*3. Bug:* Why does the default "0 days" duration create an option with `1D` duration? If I see "0 days", I expect `polls` to create an option without any time information.

---

**Step 2**

I reload the page and open the options modal again, but select "2026-03-04" (i.e., tomorrow) instead. I additionally choose timezone "America/New York" and a "1 day" duration, so `polls` previews the following:

Screenshot 2026-03-03 at 18-05-12 Umfragen – Test

Adding that option creates a `2026-03-04T05:00:00+00:00` option with `1D` duration. This is perfectly fine.

---

**Step 3**

I reload the page and open the options modal again. Default timezone is "Europe/Berlin" again. I select "2025-03-05" and disable "Whole day". The time defaults to "18:15" (which is the current local time). I also keep the default "0 days" duration.

Screenshot 2026-03-03 at 18-09-58 Umfragen – Test

*4. Suggestion:* In addition to the first suggestion above, `polls` could remember (just in the browser's local storage, per poll) what timezone was used last. The options modal should then use that timezone again instead of the default.

Adding that option creates a `2026-03-05T17:15:00+00:00` option with a `null` duration.

*5. Note:* I'd say that's expected since "0 days" was the default duration, even though I recommend changing that to "1 day", or maybe "1 hour" when the "Whole day" option is disabled. However, note that "0 days" now behaves different than before: With "Whole day" enabled it is interpreted like "1 days", but with "Whole day" disabled it's interpreted like the literal "0 days".

 *6. Bug:* This just added option isn't displayed properly, simply no option text is shown:

| Modal | Poll |
| --- | --- |
| Screenshot 2026-03-03 at 18-13-51 Umfragen – Test | Screenshot 2026-03-03 at 18-14-24 Umfragen – Test |

---

**Step 4**

I reload the page and open the options modal again. I select "2026-03-06" and timezone "America/New York". I disable "Whole day", the time now defaults to "00:30". I keep the default "0 days" duration. However, the preview was totally unexpected and super confusing at first:

Screenshot 2026-03-03 at 18-34-37 Umfragen – Test

*7. Bug:* Why is the default time "00:30"? It's 18:30 in my local "Europe/Berlin" timezone, which equals 12:30 in the selected "America/New York" timezone. The issue might be related to the following:

*8. Bug:* `polls` still uses my local "Europe/Berlin" timezone to interpret the time. This results in `pools` previewing "2026-03-06 00:30" for "Europe/Berlin", equalling "2026-03-05 18:30" for "America/New York". It should be "2026-03-06 00:30" for "America/New York" instead, which equals "2026-03-06 06:30" in "Europe/Berlin".

Adding that option creates a `2026-03-05T23:30:00+00:00` option with a `null` duration. Note 5 and bug 6 apply again.

---

**Step 5**

I reload the page and open the options modal again. I select "2026-03-07" and timezone "America/New York". I disable "Whole day", manually enter a time of "12:00" and select a "1 hour" duration. Naturally the same issues as in step 4 apply:

Screenshot 2026-03-03 at 18-50-13 Umfragen – Test

Adding that option creates a `2026-03-08T11:00:00+00:00` option with a `1H` duration. Besides bugs 7 and 8 I think that's correct.

---

**Other notes**

`composer.json` specifies a dev dependency of `"nextcloud/ocp": "dev-stable30"`. I forgot to report that last time and just happen to remember again. It didn't cause any issues now because I didn't install the dev dependencies this time. Last time I did and it conflicted with my local Nextcloud 32. This is just a heads up. I'm not sure whether this dev dependency is actually needed (no need to explain, I don't need to know), but if it is, I assume that it at least needs to be updated.

dartcafe:
Wow. Thanks for this detailed report.

## Step 1
### Zero day duration
Selecting a day with a zero duration is equal to a one day duration, if "whole day" is set. Both are interpreted as **one** day schedule, beginning at 00:00 and lasting until 24:00 (or better 23:59). This is manly to keep backwards compatibility and maybe subject of later change. 

### Selected timezone
There is only one parameter which selects the timezone ("own timezone" or "poll's timezone") and is stored in the session information. The reason, why you where faced with the local timezone is, because the initial setting is "use own timezone". If you change it in the dialog, the display setting for the poll is changed as well, assuming that the user prefers this timezone.
grafik

On reload, the setting will get reset to the default "own timezone" for the whole app. Polls does not store any information in the browser store and probably never will. Of course we could store the last setting in the users profile settings on every change. In this way we can keep the setting alive, even on reloads.

If you think this would be a solution, please open a new issue for that, because this will not be changed now.

dartcafe:
## Step 3
### Zero day|month|hours|...
As stated above, this does not select "whole day", so a zero duration means a single moment in time and therefore this option actually has no duration in contrast to the term "whole day" which represents a span by definition.

> but with "Whole day" disabled it's interpreted like the literal "0 days".

It is literally zero duration, no matter, which unit is selected.

The missing text has to be a bug indeed. I will check that!

dartcafe:
## Step 4 and 5
### timezone mismatch
Damn, I had this on time but could not reproduce that. I thought, I did not look correctly or it was because of the dev environment. 

Here the timezone seems not to be correctly applied to the display. The main information should show the chosen timezone. The other timezone is added as information. In you screenshot it is displayed the wrong way. 

Have to check and fix this.

dartcafe:
> composer.json specifies a dev dependency of "nextcloud/ocp": "dev-stable30"

That's only relevant for the dev environment, but indeed outdated, have to refresh that. As well as bringing back PHP 8.1 as supported version.

dartcafe:
Again, thanks for the support. 
What is more important: Did you observe any issues with existing polls? Any exceptions thrown?

PhrozenByte:
**Step 1, Suggestion 2 and Bug 3, titled "zero day duration"**

> Selecting a day with a zero duration is equal to a one day duration, if "whole day" is set. Both are interpreted as **one** day schedule, beginning at 00:00 and lasting until 24:00 (or better 23:59). This is manly to keep backwards compatibility and maybe subject of later change.

How do I create a option without any time information then? Because if I disable "Whole day" and choose "00:00" with a zero duration, it still points to that moment in time and thus still yields a different time in other timezones:

Screenshot 2026-03-03 at 20-07-34 Umfragen – Test

IMHO there should be a way to create a option without any time information, i.e., a option that yields e.g. "2026-03-08" without time no matter the timezone. I'd expect this precisely from selecting "Whole day" with a "0 days" duration. I guess it should end up as `iso_timestamp = 2026-03-08` (i.e., no time information) and `iso_duration = null` in the database then.

Just to make this clear: IMHO the default duration shouldn't be "0 days", but "1 days" with "Whole day" enabled, and "1 hours" with "Whole day" disabled. So, this won't be the default behaviour. Users would need to specifically create it that way, so there shouldn't be any backwards compatibility issues.

**Step 3, Suggestion 4, titled "selected timezone"**

> There is only one parameter which selects the timezone ("own timezone" or "poll's timezone") and is stored in the session information. The reason, why you where faced with the local timezone is, because the initial setting is "use own timezone". If you change it in the dialog, the display setting for the poll is changed as well, assuming that the user prefers this timezone.
>
> On reload, the setting will get reset to the default "own timezone" for the whole app. Polls does not store any information in the browser store and probably never will. Of course we could store the last setting in the users profile settings on every change. In this way we can keep the setting alive, even on reloads.
>
> If you think this would be a solution, please open a new issue for that, because this will not be changed now.

I see, thanks for the explanation. Will think about that and possibly create a new issue :+1:

**Everything else**

>  What is more important: Did you observe any issues with existing polls? Any exceptions thrown?

Unfortunately I can't say anything about that, because we don't use `polls` in production anymore (that's not `polls` fault, we just don't conduct many polls anymore :smile:). I just feel somewhat obligated to test what I can, after having voted so strongly for this feature in the past. Anything else wouldn't be fair to you, in my opinion... So, don't feel obligated to implement anything of what I've just suggested, you can absolutely disregard it, no hard feelings at all then.

> Again, thanks for the support.

You're very welcome, thanks again for your great work! :heart:

Let me know as soon as I shall do a final round of testing, will do that then.

dartcafe:
> How do I create a option without any time information then? Because if I disable "Whole day" and choose "00:00" with a zero duration

Yes. That's the week point in this. But this has been before and therefore will sty for the moment. Solving every edge case now will complicate things and delay the release. As I said in another Thread: This subject of change later.

> Just to make this clear: IMHO the default duration shouldn't be "0 days", but "1 days" with "Whole day" enabled, and "1 hours" with "Whole day" disabled.

This also. With the current logic the value will be change, from zero to 1, is someone activates the "whole day" option. And the 1 will stay even if the user switches back. 

This is a minor UX issue, which will be solved separately as well as the whole day issue. 

The main focus is not to alter existing polls. The changes can affect the outcome of existing polls. And this was tough enough.

PhrozenByte:
> Yes. That's the week point in this. But this has been before and therefore will sty for the moment. Solving every edge case now will complicate things and delay the release. As I said in another Thread: This subject of change later.

Alright, fair enough, totally up to you, as I said. I just want to recommend removing "Closes #2368" from the list of issues this PR closes then - that's the existing issue tracking that particular feature :+1:

> With the current logic the value will be change, from zero to 1, is someone activates the "whole day" option. And the 1 will stay even if the user switches back. This is a minor UX issue, which will be solved separately as well as the whole day issue. The main focus is not to alter existing polls. The changes can affect the outcome of existing polls.

I'm not sure whether we're talking about the same thing here :see_no_evil: :sweat_smile: 

I'm really just talking about the UI here, *not* the backend. Because, in my tests, if I open the options modal, the pre-selected default duration is "0 days". If I now disable "Whole day", it stays at "0 days". If I enable "Whole day" again, it still is "0 days".

What I'm suggesting is just that instead of pre-selecting "0 days" in the UI, it should pre-select "1 days" by default. If an user then disables "Whole day", that default "1 days" should automatically change to "1 hours". If an user enables "Whole day" again, it should switch back to "1 days" by default. That's it. No backend changes. It's really just about the UI here when adding new options, *not* what the backend is doing with these values later (like, treating "0 days" like "1 days" if "Whole day" is enabled - that's a whole different topic, that's why I separated them earlier). So, assuming you indeed change the default in the UI to "1 days" with "Whole day" enabled and an user manually changes that to "0 days", it still is treated the same by the backend.

Just wanted to clarify. If you still don't think that's an improvement: Fair enough, as I said, totally up to you :+1:

dartcafe:
> just want to recommend removing "Closes #2368" from the list of issues this PR closes then - that's the existing issue tracking that particular feature 👍

Why? This PR meets exactly the user story:
> As a poll participant I want to be able to view date only polls as such even if my timezone differs from the timezone of the creator, because it makes reading and filling out such a poll more easy. 

> If an user then disables "Whole day", that default "1 days" should automatically change to "1 hours". If an user enables "Whole day" again, it should switch back to "1 days" by default.

Let's avoid automatic changing of input fields, it's bad UX. The only necessity to change the unit is, when switching to "whole day" and the unit was one of hour or minute. 

Zero is the default, since the common use case will be just a moment in time, when adding a time information.

PhrozenByte:
> > just want to recommend removing "Closes #2368" from the list of issues this PR closes then - that's the existing issue tracking that particular feature 👍
> 
> Why? This PR meets exactly the user story:
> 
> > As a poll participant I want to be able to view date only polls as such even if my timezone differs from the timezone of the creator, because it makes reading and filling out such a poll more easy.

I'd say not really, no? The possibility to switch timezones is tracked in #2851 already. *Switching* timezones is different from *ignoring* timezones, which is also stated in #2368's title. The whole discussion in #2368 was about ignoring timezones IMO.

But anyway, as I said, I'm not a `polls` user anymore, and the features won't be used by me, so I shouldn't waste your free time with arguments here. Thanks again for your great work René! :heart: I'll unsubscribe for now, don't hesitate to ping me if you want a final round of testing from me :+1:

dartcafe:
> I'd say not really, no? The possibility to switch timezones is tracked in https://github.com/nextcloud/polls/issues/2851 already. Switching timezones is different from ignoring timezones, which is also stated in https://github.com/nextcloud/polls/issues/2368's title. The whole discussion in https://github.com/nextcloud/polls/issues/2368 was about ignoring timezones IMO.

The point is to decide the feature. And switching declines ignoring and vice versa. Giving the user the choice is the better approach IMHO. Further improvement is not excluded.

Anyways very much thank you for your time and effort. This helped more than you can imagine!!!

PR #4617, opened 2026-04-16
Preparation release v9.0.0

dartcafe:
Added unique index creation to the postSchemaChange and updated migration version

fix #4561

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,10 @@
 -->
 # Changelog
 All notable changes to this project will be documented in this file.
-
 ## [Unreleased]
+ - none so far
+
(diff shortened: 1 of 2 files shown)


Jerome-Herbinet:
Hello @dartcafe 
Thanks for the fix.
Can't this be backported to the 8.x.x branch (for Nextcloud 31) ?
Thanks

dartcafe:
There has to be backported more. At least #4620, #4636. And it needs to be tested very well, if it is regarding "the one customer". 

And first I have to review the changes made there.

dartcafe:
@Jerome-Herbinet I would rather prefer allowing NC31 for Polls 9.0. Less headache.

PR #4634, opened 2026-04-24
Security: Internal exception messages are returned to clients

tomaioo:
## Summary

Security: Internal exception messages are returned to clients

## Problem

**Severity**: `Low` | **File**: `lib/Controller/BaseController.php:L47`

The base controller returns raw exception messages in HTTP responses (`['message' => $e->getMessage()]`). If exception text contains internal details (query context, identifiers, stack-related hints), this can expose sensitive implementation information to attackers.

## Solution

Return generic user-facing error messages and log detailed exception information server-side only. Map known exceptions to stable, sanitized error codes/messages.

## Changes

- `lib/Controller/BaseController.php` (modified)

### DIFF
--- a/lib/Controller/BaseController.php
+++ b/lib/Controller/BaseController.php
@@ -48,7 +48,7 @@ class BaseController extends Controller {
 
 			/** @var HttpStatusCode $status */
 			$status = $e->getStatus();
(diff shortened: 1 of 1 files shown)


dartcafe:
Which messages would you classify as dangerous?

PR #4667, opened 2026-04-30
Re-enable duplicate deletion before recreating tables

dartcafe:
This step war accidentally ommitted


derStephan:
Works now, thank you very much.

PR #4814, opened 2026-06-25
fix: add salt to obescure user id further

hamza221:
## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/Helper/Hash.php
+++ b/lib/Helper/Hash.php
@@ -54,10 +54,10 @@ abstract class Hash {
 	 * This is used to create a unique identifier for users.
 	 *
 	 * @param string $userId
+	 * @param string $salt
 	 * @return string
 	 */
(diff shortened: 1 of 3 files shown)


dartcafe (approved):
Yeah. Nice. Thanks!!

hamza221:
Is it possible to backport this to Nextcloud 30 ? I can only see stable-8, no distinct stable8.6 and stable8.3

dartcafe:
Nextcloud 30 was dropped since version 8.3. I did not check, if it still works. 

But 9.1.5 will also support Nextcloud 31, which means 8 and 7 will get overridden by the next update.

This is the last tag with Nextcloud 31: https://github.com/nextcloud/polls/tree/v8.3.10

PR #4860, opened 2026-07-17
fix: purge soft-deleted entries after 12 hours as documented

solracsf:
The janitor cron purged comments, options and shares 4320 seconds (72 minutes) after their virtual deletion, while the comment states a 12 hours undo window (43200 seconds), a dropped zero, present since the purge was introduced.

Add a regression test covering both sides of the window.

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI


dartcafe (approved):
nice. I will repeat 2nd math class. ;-)

PR #5013, opened 2026-09-10
Fix typo in French localization for public link

JocelynDelalande:


dartcafe:
Translations are imported from Transifex. Any changes here will get overwritten from the import. So they have to be corrected [there](https://www.transifex.com/). 

More information is found by visiting https://nextcloud.com/translation/

